Issue #2, opened 2015-02-07
Better Errors when imports fail


chriseppstein:
See also: https://github.com/sass/node-sass/issues/651

We need to give better errors when imports from node modules fail.

Jakobo:
At a minimum, a proper node module won't install/link correctly, so when working outside of test data the problem we encountered is less likely to happen. In the future `npm init` should provide a package.json with sufficient scaffolding for tests.

chriseppstein:
@Jakobo This sounds like a good change. Do you have time to submit the patch for it?

Jakobo:
Yeah. I'll have some time today to improve it so we can drop the node-resolve dependency.

--Jakob/mobile




On Mon, Feb 9, 2015 at 9:33 AM -0800, "Chris Eppstein" <notifications@github.com> wrote:












@Jakobo This sounds like a good change. Do you have time to submit the patch for it?

chriseppstein:
FWIW, when I wrote this bug I was more concerned about handling typos like `@import "<molude_a>"` and being able to return an error like `Could not find npm module <molude_a> did you mean one of "module_a", "module_b", or "module_c"?`

chriseppstein:
Node-sass fixed this, but we removed our ability to return detect import errors in the meantime :P

PR #3, opened 2015-02-09
Improves module resolution to rely on 'module' instead of node-resolve

Jakobo:
This change updates the module_importer to rely on the node.js native 'module' implementation for resolving module dependencies. In addition to removing a dependency, this also enables inclusion of globally installed node.js modules as well as modules installed using `npm link`.

Addresses some portion of the work from #2 by returning errors as if the item had been `require()`ed in node.js parlance.

### DIFF
--- a/lib/module_importer.js
+++ b/lib/module_importer.js
@@ -1,9 +1,27 @@
 'use strict';
 
 var fs = require("fs"),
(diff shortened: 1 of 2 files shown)


chriseppstein:
:+1:

PR #7, opened 2015-02-11
Addition of coding conventions, gulp pipeline, and a code scrub

Jakobo:
This commit is uncomfortably large due to the amount of scrubbing that was done. That said, I'm really glad we're addressing this now.

* All files are now compatible with the ESLint file defined in [sass-eyeglass/eyeglass-dev-eslint](https://github.com/sass-eyeglass/eyeglass-dev-eslint).
* gulp was chosen as the build pipeline. It was easier than grunt to programatically invoke things for what we were doing

All tests are passing still, and now they all pass ESLint as well. This sets us up for CI integration.

### DIFF
--- a/build/lint.js
+++ b/build/lint.js
@@ -1,0 +1,10 @@
+var eslint = require("gulp-eslint");
+var config = require("eyeglass-dev-eslint");
+
+module.exports = function(gulp) {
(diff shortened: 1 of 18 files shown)


chriseppstein:
Great stuff.

PR #12, opened 2015-02-18
Correctly handle when an npm module is itself an eyeglass module.

chriseppstein:
This lets code in tests written for an eyeglass module work exactly like it would if it were a consumer of the module using dependencies.

### DIFF
--- a/lib/module_importer.js
+++ b/lib/module_importer.js
@@ -116,6 +116,11 @@ function makeImporter(eyeglass, sass, options, fallbackImporter) {
       }
 
       var sassDir = require(jsFile)(eyeglass, sass).sassDir;
+
+      if (!sassDir) {
+        throw new Error("sassDir is not specified in " + jsFile);
+      }
+
       var filenameSegments = [sassDir];
 
       if (relativePath) {
--- a/lib/module_importer.js
+++ b/lib/module_importer.js
@@ -2,7 +2,6 @@
 
 var fs = require("fs");
 var path = require("path");
-var resolve = require("./util/resolve");
 var efs = require("./util/files");
 var discover = require("./util/discover");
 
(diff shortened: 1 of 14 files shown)


chriseppstein:
@Jakobo please review?

lib/module_importer.js
```diff
@@ -60,6 +72,21 @@ function readFirstFile(uri, filenames, cb, examinedFiles) {
 function makeImporter(eyeglass, sass, options, fallbackImporter) {
   var importedFiles = {};
   var root = options.root;
+  var allModulesCache = {};
+
+  function getModuleByName(moduleName, dir) {
+    var allModules = allModulesCache[dir];
+    if (!allModules) {
+      allModules = discover.all(dir, true).modules;
```
Jakobo: This isn't compatible with `depth`, as the cache will be for `depth = 0` and not account for other calls to `scan`

1. remove depth
2. use `depth` of 0 as part of the `allModules` cache

lib/util/discover.js
```diff
@@ -27,7 +26,44 @@ function discover(dir) {
...
+
+    function shouldRecurse(depth) {
+      return (depth !== undefined && depth > 0) || depth === undefined;
+    }
+
+    function nextDepth(depth) {
+      return (depth !== undefined && depth > 0) ? depth - 1 : depth;
+    }
+
+    function scan(inDir, parentName, parentVersion, depth) {
```
Jakobo: Not sure `depth` is needed. Perhaps we add it back in when we have a use case? Will reduce total code.

chriseppstein: Fixed. Actually, just the `shallow` var that was in the closure was enough to get this done so the old function signature is back.

Jakobo:
:+1:

PR #18, opened 2015-03-06
declare need to clone eyeglass

sturobson:
This tripped me up. It's probably implicit, considering a user would be within the eyeglass repo when reading the other instructions. But Wondering if something more declarative like this would help?

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,12 @@ Eyeglass is using cutting edge features from node-sass that won't be
 released until node-sass 3.0. If you want to develop on or with eyeglass
 you'll need to build a custom version of node-sass for the time being:
 
+If you've not already done so:
+```
+$ git clone git@github.com:sass-eyeglass/eyeglass.git
+```
+Then:
+
 ```
 $ git clone https://github.com/sass/node-sass.git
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -9,6 +9,12 @@ Eyeglass is using cutting edge features from node-sass that won't be
...
 you'll need to build a custom version of node-sass for the time being:
 
+If you've not already done so:
+```
+$ git clone git@github.com:sass-eyeglass/eyeglass.git
+```
+Then:
+
 ```
 $ git clone https://github.com/sass/node-sass.git
```
krisbulman: makes more sense to just add ```$ git clone git@github.com:sass-eyeglass/eyeglass.git``` as a single step before this instruction than to add 5 unneeded lines.

krisbulman:
also tripped me up.

README.md
Jakobo: Agree with @krisbulman. Can you tweak it @sturobson to just have the clone line at the top of the instruction block? Thanks!

chriseppstein:
Merged. We can iterate on this if we want.

PR #21, opened 2015-03-09
Fixes #13 - Allow importing CSS files

### LINKED ISSUE #13: Allow importing CSS files
True fact: CSS files are scss files. So we can import them as long as people don't put the extension in the import (that triggers sass to treat it as a css import).

eddie-ruva:
I saw that css importing was not implemented yet as #13 describes so I decided to take so time to work on it. If you have any feedback I'll be happy to address it :smile:

### DIFF
--- a/lib/module_importer.js
+++ b/lib/module_importer.js
@@ -33,7 +33,7 @@ function readFirstFile(uri, filenames, cb, examinedFiles) {
   var filename = filenames.shift();
   examinedFiles = examinedFiles || [];
   examinedFiles.push(filename);
-  fs.readFile(filename, function(err, data) {
+  fs.readFile(filename, "utf8", function(err, data) {
     if (err) {
       if (filenames.length) {
         readFirstFile(uri, filenames, cb, examinedFiles);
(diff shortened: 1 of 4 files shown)


Jakobo:
@chriseppstein Should `.css` files follow the same rule as `.scss` / `.sass` files? (As in we look for _underscore versions of the files as well?).  There was your original Sass plugin at https://github.com/chriseppstein/sass-css-importer which seems relevant to the discussion.

Otherwise LGTM! :clap:

lib/module_importer.js
```diff
@@ -33,7 +33,7 @@ function readFirstFile(uri, filenames, cb, examinedFiles) {
   var filename = filenames.shift();
   examinedFiles = examinedFiles || [];
   examinedFiles.push(filename);
-  fs.readFile(filename, function(err, data) {
+  fs.readFile(filename, "utf8", function(err, data) {
```
pascalduez: This has the effect of returning a String instead of a Buffer previously.
Was it intended ?

chriseppstein:
@Jakobo This patch still supports css partials. The thing it doesn't allow is adding the .css extension explicitly. This would trigger Sass's css import logic.

chriseppstein:
Thanks @eddie-ruva!

PR #22, opened 2015-03-12
Assets

chriseppstein:
This is a WIP. We decided to not make a separate module for assets after all, but there was a bunch of discussion here: https://github.com/sass-eyeglass/eyeglass-assets/issues/1

### DIFF
--- a/lib/function_loader.js
+++ b/lib/function_loader.js
@@ -32,7 +32,6 @@ function checkConflicts(obj1, obj2) {
 module.exports = function(eyeglass, sass, options, existingFunctions) {
   var root = options.root;
   var functions = {};
-  hash.merge(functions, require("./functions")(eyeglass, sass));
   var modules = autoDiscoverModules(root);
 
   modules.forEach(function(m) {
--- a/lib/function_loader.js
+++ b/lib/function_loader.js
@@ -32,6 +32,7 @@ function checkConflicts(obj1, obj2) {
 module.exports = function(eyeglass, sass, options, existingFunctions) {
   var root = options.root;
   var functions = {};
+  hash.merge(functions, require("./functions")(eyeglass, sass));
   var modules = autoDiscoverModules(root);
 
   modules.forEach(function(m) {
--- a/lib/functions.js
+++ b/lib/functions.js
@@ -4,7 +4,7 @@ var hash = require("./util/hash");
 
 module.exports = function(eyeglass, sass) {
   var functions = {};
-  ["asset_url"].forEach(function(name) {
+  ["asset_url", "version"].forEach(function(name) {
     var fn = require("./functions/" + name)(eyeglass, sass);
     hash.merge(functions, fn);
   });
--- a/lib/functions.js
+++ b/lib/functions.js
@@ -1,0 +1,12 @@
+"use strict";
+
+var hash = require("./util/hash");
+
+module.exports = function(eyeglass, sass) {
+  var functions = {};
+  ["asset_url"].forEach(function(name) {
+    var fn = require("./functions/" + name)(eyeglass, sass);
+    hash.merge(functions, fn);
+  });
+  return functions;
+};
--- a/lib/functions/asset_url.js
+++ b/lib/functions/asset_url.js
@@ -1,9 +1,17 @@
 "use strict";
 
+var unquote = require("../util/unquote");
+
 module.exports = function(eyeglass, sass) {
   return {
-    "asset-url($relative-path)": function(relativePath, done) {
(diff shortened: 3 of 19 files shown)


lib/function_loader.js
```diff
@@ -32,6 +32,7 @@ function checkConflicts(obj1, obj2) {
 module.exports = function(eyeglass, sass, options, existingFunctions) {
   var root = options.root;
   var functions = {};
+  hash.merge(functions, require("./functions")(eyeglass, sass));
```
Jakobo: We probably want to merge our "protected" functions at the end. It would also allow us to offer better errors/warnings if it's a collision vs overriding one of eyeglass' native methods

lib/functions.js
```diff
@@ -0,0 +1,12 @@
+"use strict";
+
+var hash = require("./util/hash");
+
+module.exports = function(eyeglass, sass) {
+  var functions = {};
+  ["asset_url"].forEach(function(name) {
+    var fn = require("./functions/" + name)(eyeglass, sass);
```
Jakobo: I really like this pattern. Very easy for us to add functions later. :+1:

lib/functions/asset_url.js
```diff
@@ -0,0 +1,9 @@
+"use strict";
+
+module.exports = function(eyeglass, sass) {
+  return {
+    "asset-url($relative-path)": function(relativePath, done) {
+      done(sass.types.String("url(" + relativePath.getValue() + ")"));
```
Jakobo: Just a note for later: We should probably figure out the `wait.for` solution so that eyeglass custom functions are node-standard callbacks by default. This means making `done()` instead read as

`done(null, sass.types.String("url(" + relativePath.getValue() + ")"))`

This way when you're in JS, it feels like proper node. If you return an `Error` object, we can class the message into `sass.types.Error` for them.

Jakobo:
:tada: Thanks @chriseppstein for starting on this. It looks really clean. I'll be opening a separate thread around node-style callbacks in eyeglass, so that things in node feel node-like.

lib/function_loader.js
chriseppstein: Huh, I actually had the opposite inclination. Since this isn't a plugin, being able to override it seemed like a good usecase.

Jakobo: That seems reasonable. Dangerous, but reasonable. We probably need some way to get at the old function for passthrough purposes. Doesn't block this PR

lib/functions/asset_url.js
chriseppstein: I like this, tracked via https://github.com/sass-eyeglass/eyeglass/issues/23

lib/function_loader.js
chriseppstein: I'm kind of an anarchist w.r.t. software design decisions like this.

lib/functions.js
chriseppstein: Yeah! I do hope to keep this to a small set of functions, though.

lib/function_loader.js
Jakobo: Totally okay. Good to know so we can keep it philosophically consistent.

chriseppstein:
Rebased against master. refactored eyeglass into an object api. Added the basic application api for adding assets to your application from within the project directory or external to it. Next I will work on referring to assets in eyeglass modules.

chriseppstein:
I've been thinking that the asset-url function should support a similar syntax to the importer. That is:

`asset-url("<foo>/foo.png")` if you want to reference an image from a module. and `asset-url("foo.png")` would only return assets from your application itself. Thoughts? @Jakobo @eoneill @scottdavis @ericam

ericam:
I like that consistency - though I've heard some concerns about the syntax (*cough* @aaronagray) — primarily because `<foo>` is standard pseudo-code for "replace this (including the brackets) with your own code. So documentation `<foo>/foo.png` seems to imply `bar/foo.png` rather than `<bar>/foo.png`. That may be more of a documentation issue than an actual syntax problem, but it's worth noting.

eoneill:
I like `asset-url("<foo>/foo.png")` and keeping with the convention we've already established. This removes any ambiguity.

You could also easily support transitive dependency scope and offer appropriate errors when attempting to access a module's assets that are out of scope.

chriseppstein:
@ericam @aaronagray, That's a good point. In a certain sense `<foo>` means substitute the full path to the module right here. But it's done by the code instead of by the user. I think we'd have the same ambiguity with `[foo]`. So I'm open to suggestions about what the node module reference syntax should look like. I opened a new issue to track this going forward: https://github.com/sass-eyeglass/eyeglass/issues/26

PR #24, opened 2015-03-18
Removes need to differentiate between sync/async custom functions

Jakobo:
This creates a single pattern (like importers) for custom functions. This code
can be removed when `node-sass` has similar support. For now, we have an
additional `.gyp` dependency when using this patch. The `deasync` library
allows us to create a blocking `while()` loop and drain the event queue. This
pattern is a bit better for our application than using a `node-fiber`, since
we want to not replumb the entire custom function flow.

Custom functions w/ this patch can either call `done()` (always the last
argument), or return a value synchronously.

Includes test.

### DIFF
--- a/lib/eyeglass-exports.js
+++ b/lib/eyeglass-exports.js
@@ -1,0 +1,11 @@
+"use strict";
+
+var path = require("path");
+
+module.exports = function(eyeglass, sass) {
+  var opts = {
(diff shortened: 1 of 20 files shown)


Jakobo:
Ref: http://stackoverflow.com/questions/13448374/how-to-sleep-the-thread-in-node-js-without-affecting-other-threads

//cc @chriseppstein @eoneill

PR #27, opened 2015-03-20
PROPOSAL: Conceptual model of non angle bracket based imports

Jakobo:
This is a proof of concept for using pure node.js syntax for imports. The custom importer has been modified to pass modules through silently, letting an unresolved eyeglass module defer to relative imports and finally the native node-sass importer.

Benefits of this implementation:
- `@import` will resolve identically to `require()` from within the sass file
- If an npm style `require` path does not work, it will fall back to Sass' resolution system
- This avoids any additional syntax. No `<foo>`, no "npm!foo" or other syntax to request the node resolution system

Drawbacks of this implementation:
- We no longer can tell the intent of an import. It's ambiguous if the intent was to import an eyeglass module or do a traditional Sass import. This means we lose a layer of error reporting. This is a deficiency in `node-sass`, as errors from the prior Importer are not able to be passed down through the previousImporter function.
- There is a learning curve for someone completely unfamilliar with the npm ecosystem.
- You risk a collision if you have a directory called "foo" and then you went and installed the "foo" npm module (which happened to also be an eyeglass module). Since we are encouraging people to name their modules `eyeglass-*`, I don't forsee this being a serious drawback.

### DIFF
--- a/lib/module_importer.js
+++ b/lib/module_importer.js
@@ -5,8 +5,6 @@ var path = require("path");
 var efs = require("./util/files");
 var discover = require("./util/discover");
 
-var IMPORT_REGEX = /^<([^>]+)>(?:\/(.+))?$/;
-
 /*
  * All imports use the forward slash as a directory
  * delimeter. This function converts to the filesystem's
@@ -105,24 +103,15 @@ function makeImporter(eyeglass, sass, options, fallbackImporter) {
 
   return function(uri, prev, done) {
     var isRealFile = fs.existsSync(prev);
-    var match = uri.match(IMPORT_REGEX);
+    var fragments = uri.split("/");
+    var moduleName = fragments[0];
+    var relativePath = fragments.slice(1).join("/");
+    var pkgRootDir = isRealFile ? packageRootDir(path.dirname(prev)) : root;
+    var jsFile = getModuleByName(moduleName, pkgRootDir);
 
-    if (match) {
-      // This is an import of the form "<node_module>/some/file"
-      var moduleName = match[1];
-      var relativePath = match[2];
-      var pkgRootDir = isRealFile ? packageRootDir(path.dirname(prev)) : root;
-      var jsFile = getModuleByName(moduleName, pkgRootDir);
-
-      if (!jsFile) {
-        console.error("No Eyeglass module named '" + moduleName +
-                      "' could be found while importing '" + uri + "'.");
-        done({});
-        return;
-      }
-
+    if (jsFile) {
+      // is eyeglass module
       var sassDir = require(jsFile)(eyeglass, sass).sassDir;
-
       if (!sassDir) {
         throw new Error("sassDir is not specified in " + jsFile);
       }
@@ -145,10 +134,8 @@ function makeImporter(eyeglass, sass, options, fallbackImporter) {
           importOnce(data, done);
         }
       });
-
     } else if (isRealFile) {
-      // This is a sass file that is potentially relative to the
-      // previous import.
+      // relative file import, potentially relative to the previous import
       var f = path.resolve(path.dirname(prev), makeFsPath(uri));
       readAbstractFile(uri, f, function(err, data) {
         if (err) {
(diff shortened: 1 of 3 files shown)


lib/module_importer.js
```diff
@@ -105,24 +103,15 @@ function makeImporter(eyeglass, sass, options, fallbackImporter) {
...
-
-    if (match) {
-      // This is an import of the form "<node_module>/some/file"
-      var moduleName = match[1];
-      var relativePath = match[2];
-      var pkgRootDir = isRealFile ? packageRootDir(path.dirname(prev)) : root;
-      var jsFile = getModuleByName(moduleName, pkgRootDir);
-
-      if (!jsFile) {
-        console.error("No Eyeglass module named '" + moduleName +
```
Jakobo: This is the eyeglass specific error reporting we lose.

lib/module_importer.js
```diff
@@ -105,24 +103,15 @@ function makeImporter(eyeglass, sass, options, fallbackImporter) {
...
-        return;
-      }
-
+    var fragments = uri.split("/");
+    var moduleName = fragments[0];
+    var relativePath = fragments.slice(1).join("/");
+    var pkgRootDir = isRealFile ? packageRootDir(path.dirname(prev)) : root;
+    var jsFile = getModuleByName(moduleName, pkgRootDir);
+
+    if (jsFile) {
```
Jakobo: By testing positive for `jsFile` instead of negative, we can slip through to the `isRealFile` path or the native Sass importer.

eoneill:
Overall, I like this, but I question whether or not we should _start_ here. The syntax is cleaner, but I think the potential for collision could burn people.

I think I'd still rather do something like `@import "<foo/bar>"` and be explicit that this is pointing externally. We have the ability to relax this in the future if we decide it's unnecessary, but it'll be harder to do the reverse. Having an eyeglass specific syntax also affords us the ability to deprecate this pattern and provide useful warnings.

My other issue with this approach is that it doesn't quite work the same way node's `require` does (without fundamentally breaking Sass' importer).

e.g. in node...
```js
require('module_a'); // always guaranteed to resolve to an external module
require('./module_a'); // only way to import a local file
```
however in Sass...
```scss
// these are likely the same file, though there are other import rules to consider
@import "module_a";
@import "./module_a";
```
I'm also not confident that everyone is going to be naming their modules `eyeglass-*`.

Jakobo:
`@import "require('module_a')"` Then everyone wins! :laughing: 

Regarding the other concern about paths, the fallback importer resolves this correctly. If `module_a` isn't a valid eyeglass module, it hands off to `node-sass` (and thus `libsass`), which will look for `module_a/index.scss...` and all the standard Sass path resolutions.

I just don't think it's worth the extra syntax to get clever. :smile: 

People will name their public eyeglass stuff `eyeglass-*` because otherwise it will be very difficult to find in the npm ecosystem. We know this historically from many other plugin systems that were developed.

chriseppstein:
> People will name their public eyeglass stuff eyeglass-* because otherwise it will be very difficult to find in the npm ecosystem.

Yes, but In your package.json you can specify an eyeglass name that is used for importing, etc. and this name can be nicer. E.g. If there's a package `eyeglass-susy` the import could be made to be `@import "<susy>";` by setting the eyeglass name to susy.

chriseppstein:
Honestly, namespace collisions don't worry me, people can rename their directory if they need to.

lib/module_importer.js
eddie-ruva: @Jakobo shouldn't we be able to show this error in the last case of this [if-else block](https://github.com/sass-eyeglass/eyeglass/blob/master/lib/module_importer.js#L167)?

Jakobo: We can, but we won't know if they intended the import as a node style or not. Still, I think that any error is better than none and will add the suggestion and rebase :)



—
Jakob Heuser

eddie-ruva: Oh that's true I forgot that it could happen. The rest of the proposal looks really nice :+1:

chriseppstein:
I've rebased, fixed broken tests, updated the docs and landed this on master.

PR #30, opened 2015-04-15
importer: Only use done() callback method

saper:
In node-sass 3.x last "done" parameter is
always passed to the custom importer.

It should be not necessary to check if
done() has been provided.

Should-Solve: https://github.com/sass/node-sass/issues/849

### DIFF
--- a/lib/module_importer.js
+++ b/lib/module_importer.js
@@ -123,19 +123,10 @@ function makeImporter(eyeglass, sass, options, fallbackImporter) {
 
   function importOnce(data, enabled, done) {
     if (enabled && importedFiles[data.file]) {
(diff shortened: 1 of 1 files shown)


eddie-ruva:
@saper I think it looks good, I would just also delete readAbstractFileSync and readFirstFileSync since they are not used anymore.

chriseppstein:
I've merged this to master (rebased). I added [more tests around synchronous rendering](https://github.com/sass-eyeglass/eyeglass/blob/master/test/test_sync.js) to be extra sure :)

chriseppstein:
@eddie-ruva I removed the unused functions as well.

chriseppstein:
@saper Thank you for the patch.

saper:
:checkered_flag:

PR #35, opened 2015-05-20
semver support at runtime for eyeglass modules

Jakobo:
This change adds semver support to eyeglass modules, avoiding the problems
typically associated with peerDependencies. Instead, an eyeglass module can
add to their package.json:

```
{
  "eyeglass": {
    "engine": "SEMVER_VALUE"
  }
}
```

And set `SEMVER_VALUE` to a semver compatible string. When eyeglass wraps the
sass options and eyeglass modules are loaded, the `eyeglass.engine` values are
checked against the currently running eyeglass version. If an incompatible
engine is found, the user has the option to stop the node-sass process from
continuing.

To force a hard stop, set `eyeglassStrictVersions: true` in your node-sass
options.

To silence all messages, set `eyeglassStrictVersions: false` in your node-sass
options.

By default, warnings are thrown instead of halting, and the value of
`eyeglassStrictVersions` is `"warn"`.


chriseppstein:
Ok. Bikeshed comments:

* I don't like the name `eyeglassStrictVersions` for what this does. To me, it means that it will check versions of plugins. I think it should be `allowIncompatiblePlugins` and still default to warn.
* I think "engine" should be "requireEyeglassVersion".

Other comments:

* Should also check against the node-sass/libsass version since we dependency inject both into the plugins?

```json
{ "eyeglass": {
   "requires-version": {
    "node-sass": "SEMVER_VALUE",
    "libsass": "SEMVER_VALUE",
    "eyeglass": "SEMVER_VALUE"
   }
  }
}

**Comment by chriseppstein** [comment]
Here's something I'm worried about. Someone has module that they want to allow even though it declares itself incompatible (they tested and it still works for them). So they turn warnings off. But then they install a new module and it has problems but they don't see a warning.

I'm wondering if we should limit silencing warnings only by specifying a new semver compat for that plugin. `allowedVersionOverride: { "eyeglass-math": "^2.0.0"}` or even something simpler: `forcePluginLoad: ["eyeglass-math"]`

File: lib/module_importer.js [hunk 1]
```diff
@@ -5,13 +5,6 @@ var path = require("path");
 var efs = require("./util/files");
 var discover = require("./util/discover");
 
-// Returns whether a file exists.
-function existsSync(file) {
-  // This fs method is going to be deprecated
-  // but can be re-implemented with fs.accessSync later.
-  return fs.existsSync(file);
-}
-
```
chriseppstein: I fixed this warning by actually using it in another branch.

lib/semver_checker.js
```diff
@@ -0,0 +1,62 @@
...
+
+  var incompatible = {};
+  var noEngine = [];
+  var hasErrors = false;
+  var hasWarnings = false;
+
+  strictMode = (typeof strictMode === "undefined") ? "warn" : strictMode;
+
+  modules.forEach(function(m) {
+    if (!m.eyeglass || !m.eyeglass.needs) {
```
chriseppstein: In the pull request description it was `eyeglass.engine` but all the code seems to be using `needs` which I like better.

package.json
```diff
@@ -24,7 +24,8 @@
     "eyeglass-module"
   ],
   "eyeglass": {
-    "exports": "lib/eyeglass-exports"
+    "exports": "lib/eyeglass-exports",
+    "needs": "*"
```
chriseppstein: I feel like it should be a warning to have `*` as a version requirement. It's super dangerous to not even require a specific major version.

test/testutils.js
```diff
@@ -68,7 +69,10 @@ module.exports = {
     var lines2 = string2.split("\n");
     assert.equal(lines1.length, lines2.length, "Number of lines differ.");
     for (var lineNumber = 0; lineNumber < lines1.length; lineNumber++) {
-      assert.equal(lines1[lineNumber], lines2[lineNumber], "Line #" + lineNumber + " differs: " + lines1[lineNumber] + " != " + lines2[lineNumber]);
+      assert.equal(lines1[lineNumber], lines2[lineNumber], [
+        "Line #", lineNumber, " differs: ",
+        lines1[lineNumber], " != ", lines2[lineNumber]
+      ].join(" "));
```
chriseppstein: I also addressed these line length warnings. you'll probably get conflicts when you rebase/merge.

chriseppstein:
Overall, this is a great patch. I have some minor nits and questions I've asked above. Once we sort that out, let's merge it in!

lib/module_importer.js
Jakobo: k. will rebase and add back

lib/semver_checker.js
Jakobo: Yeah, per our conversation, I updated it to `needs` and `strictModuleVersions`

package.json
Jakobo: Happy to add that after the first support lands. I totally agree that `*` is stupid, but as a plugin developer writing code explicitly for other people, maybe we expect better?

test/testutils.js
Jakobo: rebased. thanks

Jakobo:
@chriseppstein Made all the change suggested from the diffs above. This would probably bump our eyeglass version to signal that warnings will now be thrown. (but it won't break anyone's code)

chriseppstein:
Merged in after rebasing.

PR #36, opened 2015-05-21
Documentation

Jakobo:
The project's big enough to need documentation. In the #eyeglass channel on
Slack, we agreed to put everything into the code, as a wiki falls out of date
quickly. This contains documentation for the following:

* Getting Started
* Installing & using eyeglass modules
* Creating an eyeglass module
* Defining custom functions
* Troubleshooting
* A "Why" doc

### DIFF
--- a/docs/eyeglass_modules/index.md
+++ b/docs/eyeglass_modules/index.md
@@ -1,0 +1,102 @@
+# Writing an eyeglass Module
+
+Writing an eyeglass module is a great way to both learn eyeglass and take common Sass code and make it reusable and managable from a dependency perspective. If you want to jump right in, we recommend referencing the [sample module](https://github.com/sass-eyeglass/eyeglass-sample) or using the [Yeoman generator](https://github.com/sass-eyeglass/generator-eyeglass).
+
+# eyeglass Folder Structure
+
+An eyeglass module consists of (at a minimum) the following:
+
+```
+/my-eyeglass-module <directory>
+  /package.json
+  /eyeglass-exports.js
+  /scss <directory>
+```
+
+eyeglass modules are node.js modules, and in order to work with npm, they need a `package.json` file. You can quickly create one using the `npm init` command and accept the default options. While we won't go into all the details of `package.json` here, the [npm package.json documentation](https://docs.npmjs.com/files/package.json) is really complete.
+
+If you don't like `eyeglass-exports.js` and want to be more node-like, feel free to change it to `index.js`. You'll just set the [main property](https://docs.npmjs.com/files/package.json#main) in `package.json` accordingly.
+
+The `scss` directory is optional, but many common eyeglass modules provide Sass files, so it might be good to make the directory ahead of time.
+
+# Adding "eyeglass" to package.json
+
+The next step is to add the `eyeglass` section to package.json. This gives eyeglass important information about your module, including compatibility information and key overrides.
+
+```js
+// ...
+  "eyeglass": {
+    "needs": "<eyeglass version>",
+    "exports": "<optional: alternate exports>"
+  }
+```
+
+Currently, the `eyeglass` section contains two properties, `needs` and `exports`.
+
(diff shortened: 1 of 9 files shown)


docs/index.md
```diff
@@ -0,0 +1,28 @@
...
+  * [eyeglass support for existing Sass projects](eyeglass_modules/index.md#bringing-eyeglass-to-existing-projects)
+* [Why eyeglass?](why_eyeglass.md)
+
+# Introduction: How eyeglass works
+
+When we set out to create eyeglass, we wanted to create a system that took the best parts of the Sass language and community, and sprinkled in a really good dependency management system. While it might have been possible to create a dependency management system, repository, and website all on our own `node-sass` development was taking off. With Custom Functions implemented, there was no reason npm couldn't become the ecosystem for distributing and consuming Sass modules for `node-sass`.
+
+To avoid creating any additional tools, we **wrap node-sass options** with an eyeglass constructor. This allows eyeglass to scan your `node_modules`, looking for any additional eyeglass modules that are installed. Before compiling, all eyeglass modules are added to the include path, and all Custom Functions are added to the `functions` object of `node-sass`. In many ways, eyeglass automates the module discovery and option decoration.
```
chriseppstein: Actually, eyeglass modules are not added to the include path. There's a custom importer.

docs/why_eyeglass.md
```diff
@@ -0,0 +1,24 @@
+# Why eyeglass?
+eyeglass fills a need for `node-sass` users: access to the rich Sass ecosystem using well-tested node.js development tools. The Sass community has numerous plugins, including [compass](http://compass-style.org), [Susy](http://susy.oddbird.net), [Bourbon](http://bourbon.io), and [many more](http://www.sache.in). If eyeglass is successful, many of these awesome frameworks will become available to `node-sass` users, accessible from a simple `npm install`.
```
chriseppstein: "If eyeglass is successful" oh ye of little faith.

How about, "By standardizing on eyeglass"

docs/why_eyeglass.md
```diff
@@ -0,0 +1,24 @@
...
+More involved build systems such as Broccoli (and thus ember-cli) contain their own asset pipeline logic. The introduction of eyeglass-* plugins in these cases were to ensure any additional assets an eyeglass module might create are properly captured in the system.
+
+# Why Is eyeglass Handling Assets?
+Assets (sprites, fonts, images, etc) are inherently complex items. When you build a CSS file with `node-sass`, you may want to change local paths into CDN locations, move files around, or otherwise create new items from your raw sources. If we didn't put this into eyeglass' core, we would have made an `eyeglass-assets` module. However, there's so much configuration in the assets themselves (based on our experience with compass), that we felt asset management should be part of eyeglass' core behaviors.
+
+Key discussions on the topic:
+* [Asset Installation](https://github.com/sass-eyeglass/eyeglass/issues/15)
+* [General Design](https://github.com/sass-eyeglass/eyeglass-assets/issues/1)
+
+# Why Not Just Use PostCSS?
```
chriseppstein: I don't see a point to discussing postcss in this documentation unless we're showing how to make eyeglass into a postcss build step.

docs/eyeglass_modules/index.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+* **sass.types.TYPE**: All things we pass back to `node-sass` have a `TYPE`. This tells Sass what we're dealing with, and also allows for cool Sass features like applying math to px/em/rem and having it _just work_. A full list is in the [node-sass documentation](https://github.com/sass/node-sass#functions--v300---experimental), but the three most common are `sass.types.Number(amount, unit)`, `sass.types.String(value)`, and `sass.types.Color(red, green, blue, alpha)`. If you don't want to return anything, you can return `sass.types.Null()`.
+* **name.getValue()**: Since we asked for `$name` in the **signature** and put quotes around it, `node-sass` knows to make this a String for us. Calling `.getValue()` returns the value as a String. For a String, that's not very exicitng. However, for values like "24px", `node-sass` will automatically split the result (and provide `.getValue()` and `.getUnit()`).
+
+That's a lot for one little function! Remember:
+
+1. A Custom Function consists of a **signature** and a **callback**
+2. Write the **signature** as if you were calling the function yourself. Include sensible defaults
```
chriseppstein: `Write the **signature** as if you were defining the function in a sass file`

chriseppstein:
This is a good first step. Eventually, I'd like to have a proper website that is designed and whatnot. At that point we'll probably introduce a static site compiler.

chriseppstein:
Feel free to merge this when the edits are done.

docs/why_eyeglass.md
Jakobo: Happy to remove it. However, I think this is an important question. There's already a "tool" for "css" in "npm". How's this any different / why not just use the other one?

PR #37, opened 2015-05-22
Path Fixes

Jakobo:
The paths were being scanned incompletely. We were only checking for some of the possible include scenarios, and hadn't accounted for nested directories that shared names identical to the npm module name.

This fix rewrites the file util to be a bit smarter about path expansion and reduces the number of times we're writing forEach loops.

Adds tests to catch the nested-"susy" use case.


chriseppstein:
merged. Thank you!

Issue #33, opened 2015-05-08
Make @import empty


idflood:
I can't make eyeglass work, maybe it's something related to recent libsass/node-sass updates. My setup involve gulp and sass files but I tried to convert the files to scss too without success.

For reference my gulpfile contains something like this:

    var eyeglass = new Eyeglass(sassOptions, sass.compiler);
    eyeglass.assets("assets/images", "images", "public/images");
    eyeglass.assets("assets/fonts", "fonts", "public/fonts");

    gulp.src('src/*.sass')
      //.pipe(sass(eyeglass.sassOptions()))
      .pipe(sass(sassOptions))
      .pipe(gulp.dest('dist'));

If a sass file contains something like this:

    @import "modules/my_module"
    body
      background: red

The result looks like this:

    body{background: red;}

In fact the problem exist as soon as I create a new instance of `Eyeglass`.

idflood:
The problem might be in gulp-sass instead since if I try with node-sass it works as expected.

chriseppstein:
@Jakobo did you ever figure out how to make gulp integration work?

chriseppstein:
We had some problems with gulp because we need some hook at the start of every new file so we can reset the import-once cache.

broccoli integration is working great though. https://github.com/sass-eyeglass/broccoli-eyeglass

idflood:
I'm not sure this is relevant but gulp-sass use `options.data` instead of `options.file`:
https://github.com/dlmanning/gulp-sass/blob/master/index.js#L33

In module_importer.js the `isRealFile` is always false with gulp and a console.log(prev) just after that return `stdin`.

chriseppstein:
We have a [gulp integration guide](https://github.com/sass-eyeglass/eyeglass/blob/master/docs/integrations/gulp.md). The main thing you need from that is the `eyeglass.enableImportOnce = false` part.

idflood:
I still can't make it to work with eyeglass : /

I've created a test repository showing the problem here: https://github.com/idflood/gulp-eyeglass-test

Eyeglass functions and import works as expected:
- source: https://github.com/idflood/gulp-eyeglass-test/blob/master/src/test4.scss
- output: https://github.com/idflood/gulp-eyeglass-test/blob/master/dist/test4.css

But I still don't have any luck with normal `@import`:
- source: https://github.com/idflood/gulp-eyeglass-test/blob/master/src/test1.scss
[...]

PR #39, opened 2015-05-30
Update gulp.md

pnuzhdin:
I have no idea why it works but it's required for the regular @import

See:
https://github.com/sass-eyeglass/eyeglass/issues/33
https://github.com/idflood/gulp-eyeglass-test/pull/1

### DIFF
--- a/docs/integrations/gulp.md
+++ b/docs/integrations/gulp.md
@@ -6,7 +6,7 @@ var gulp = require("gulp");
 var sass = require("gulp-sass");
 
 var eyeglass = require("eyeglass")({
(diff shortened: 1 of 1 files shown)


Jakobo:
@pnuzhdin do you mind adding a sentence or two around line 3 about why we need to include a fallback importer in the gulp implementation?

/cc @chriseppstein

pnuzhdin:
@Jakobo I can just repeat my PR description, I have no idea but it's required for the regular @import. I've investigated how eyeglass works and saw the same code has been called here but it doesn't lead to any effect for the regular @import.

PR #48, opened 2015-07-09
Assets

chriseppstein:
Ok. Assets are here and they are a little different than we designed initially. The biggest change is because node-sass does not tell custom Sass function (in javascript) what Sass file it was called from. Instead, we have implemented as asset registration system based on the importer mechanism and I actually really like this. There is a global map `$eg-registered-assets` which lists all the assets that you import. There is a special import `@import "assets"` which will register the assets for the current application or eyeglass module. You can also `@import "module-name/assets"` for any direct dependency module that the current sass file belongs to.

Assets that are exposed by importing are registered in applications by calling `eyeglass.assets.addSource(...)` and in plugins by setting the assets property in eyeglass-exports to `eyeglass.assets.export(...)`. Export is basically an alias for addSource, I just like how it sounded for the plugin use case.

Code review please.

Ping: @Jakobo @eoneill @scottdavis 

Remaining tasks for assets:

- [ ] Documentation
- [x] make asset registration functions (in addition to the mixins)
- [ ] expose asset resolution public js api for eyeglass map
- [ ] Bring up test coverage so CI pass


sass/assets.scss
```diff
@@ -1,3 +1,58 @@
...
+// over module assets, that is, "foo/bar.png" in the application will
+// take priority over "bar.png" in the "foo" module.
+// The values of this map is another map of source urls to the full
+// path on disk where the asset resides.
+$eg-registered-assets: () !default;
+
+// Registers many assets for a module at once.
+// @param $module string The name of the module to which these assets belong.
+// @param $assets map<string => string> maps source urls to the full path of the asset.
+@mixin register-asset($module, $url, $path, $uri: null) {
```
Jakobo: For convention, should these functions all be `asset-*`? I know that'd make the `register-assets` a lot wordier, but not sure we want to be in multiple namespaces.

lib/assets.js
```diff
@@ -0,0 +1,244 @@
...
+function resolveAsset(sass, sassUtils, registeredAssetsMap, relativePathString) {
+  registeredAssetsMap = sassUtils.handleEmptyMap(registeredAssetsMap);
+  sassUtils.assertType(registeredAssetsMap, "map");
+  sassUtils.assertType(relativePathString, "string");
+
+  var relativePath = unquote(relativePathString.getValue());
+  var registeredAssets = sassUtils.castToJs(registeredAssetsMap);
+
+  var appAssets = registeredAssets.coerce.get(null);
+  if (appAssets) {
```
scottdavis: Isn't this always going to be true?

scottdavis: since you are coercing a null? its not going ot be false or undefined

lib/assets.js
```diff
@@ -0,0 +1,244 @@
...
+    // XXX sassUtils.assertType(appAssets, "map");
+    var appAsset = appAssets.coerce.get(relativePath);
+    if (appAsset) {
+      return appAsset;
+    }
+  }
+
+  var segments = relativePath.split("/");
```
scottdavis: isn't `/` going to break on windows?

lib/assets_importer.js
```diff
@@ -0,0 +1,140 @@
...
+
+  function getModuleByName(moduleName, dir) {
+    var allModules = allModulesCache[dir];
+    if (!allModules) {
+      allModules = discover.all(dir, true, dir === root).modules;
+      allModulesCache[dir] = allModules;
+    }
+
+    for (var i = 0; i < allModules.length; i++) {
+      if (moduleName === allModules[i].eyeglassName) {
```
scottdavis: should assign `allModules[i]` to a variable for scoped reuse

sass/assets.scss
chriseppstein: I guess they could be `asset-register` and `asset-register-all`.

lib/assets.js
chriseppstein: if there are no registered app assets, then Sass will return a null which is coerced back to a js null which is falsy.

lib/assets.js
chriseppstein: Yes. We need to fix this.

matiassingers:
```js
var segments = relativePath.split(path.sep):
```

Should work, but you might want to take another look at this section of figuring out the module name.

----

Perhaps split it out into a separate module.

Issue #47, opened 2015-06-29
(Grunt) includePaths not working


FransTwisk:
When using eyeglass via Grunt and configuring it as explained in the example, the includePaths don't seem to work. I'm using node-sass@3.2 and grunt-sass@1.0, currently configured as:

```
...
var neat = require('node-neat').includePaths;
var paths = ['./sass/'].concat(neat); // this will also add bourbon

grunt.initConfig({
    ...
    sass: {
      options: require("eyeglass")({
        includePaths: paths,
        sourceMap: true,
        outputStyle: 'compressed'
      }).sassOptions(),
      files: {
        '<%= clientCssOutputPath %>main.css': '<%= clientSassPath %>main-extend.scss',
        '<%= clientCssOutputPath %>ie.css': '<%= clientSassPath %>ie-extend.scss',
      }
});
...
```
I left out some variables and code, but all folders are correct. Getting the following error:
```
Could not import main from any of the following locations: ...
```

The locations it looks in never include the includePath folders. This works if I don't initialise eyeglass via:
```
...
options: {
    includePaths: paths,
    sourceMap: true,
    outputStyle: 'compressed'
},
...
```

Jakobo:
Is the `main` it's trying to import coming from neat/bourbon?

FransTwisk:
No, it's coming from main-extend.scss. It's located in the root of ./sass/

alanhogan:
+1 

I wrote two small example scripts, one using node-sass directly, and one using `eyeglass.sassOptions()` as per the project README.

When using node-sass directly, `includePaths` works perfectly…

When using Eyeglass, it never makes it; the error message makes it clear it isn’t looking in `includePaths`.

This issue is probably the root cause of https://github.com/sass-eyeglass/broccoli-eyeglass/issues/13.

alanhogan:
If your issue is the same as mine, note that I think I have fixed this. I'll be making a PR shortly!

alanhogan:
@FransTwisk Would you mind seeing if this patch solves your problem? I haven’t tried Gulp. https://github.com/sass-eyeglass/eyeglass/pull/51/files

JohnRiv:
I upgraded to 0.6.1 and this pull request appears to be causing an issue now with grunt-sass. includePaths is coming through as an array now, so the call to `split` results in `Fatal error: Cannot read property 'split' of undefined`

[...]

PR #51, opened 2015-08-18
Use includePaths when importing sass files

alanhogan:
This adds functionality to the module importer to look for sass files in locations specified in `options.includePaths`

Should solve #47 and https://github.com/sass-eyeglass/broccoli-eyeglass/issues/13

Please take a look at the line where I had to turn a string into an array. That seemed odd as includePaths is supposed to be an array per https://github.com/sass/node-sass documentation, but it was needed because paths are coming in as strings separated by colons (on my Mac, at least).

Note also I’m not 100% this is the best fix for the problem. `node-sass` docs talk about returning `sass.NULL` when an importer does not want to engage for the current `uri`. There may be a simpler fix involving falling back to default `node-sass` import behavior if no eyeglass module is found, instead of using our own code to locate the file. (I briefly and naïvely experimented with this but couldn’t get it to work, but I think this may be because I don’t have a full understanding of the entire `importer`subsystem)

The code in `makeImporter` almost reads like this was the intent, but `isRealFile` checks for the existence of the `prev`ious file, not the current file; I don’t understand this behavior.


alanhogan:
@chriseppstein  @Jakobo I’ve added a test to this PR. It passes in this branch and fails in master, as expected :) 

Let me know if there is anything else I can do to help get this merged in.

Oh and a big thanks for all your work so far. I, for one, truly appreciate everything you’ve done and been doing for the Sass community.

lib/module_importer.js
```diff
@@ -32,8 +32,17 @@ function readFirstFile(uri, possibleNames, filenames, cb) {
...
+  var possibleNames = efs.expandFileName(uri, origin, location),
+    thisLocation;
+
+  // Useful for further extension/debugging of the algorithm:
+  // console.log("Note: uri, origin, location: ", uri, origin, location);
+
+  for (var i in includePaths) {
+    thisLocation = path.resolve(location, includePaths[i]);
+    possibleNames = possibleNames.concat(efs.expandFileName(uri, origin, thisLocation));
+  }
```
eoneill: can we use `.forEach` to make this slightly easier to read...

```js
includePaths.forEach(function(includePath) {
  var includeLocation = path.resolve(location, includePath);
  possibleNames.concat(efs.expandFileName(uri, origin, includeLocation));
});
```

lib/module_importer.js
```diff
@@ -56,6 +65,7 @@ function makeImporter(eyeglass, sass, options, fallbackImporter) {
     var relativePath = fragments.slice(1).join("/");
     var pkgRootDir = isRealFile ? importUtils.packageRootDir(path.dirname(prev)) : root;
     var jsFile = importUtils.getModuleByName(moduleName, pkgRootDir);
+    var includePaths = options.includePaths.split(path.delimiter); // surprised this is not already an array, coming from broccoli-eyeglass
```
eoneill: consider...

```js
var includePaths = Array.isArray(options.includePaths) ? options.includePaths : options.includePaths.split(path.delimiter);
```
so we're future proofed if down the road it starts coming in as an array rather than a string

eoneill:
Code changes look fine, but I don't really know enough about how the importer is actually working under the hood. So I'll wait to see if there's any other feedback.

Thanks for digging into this!

lib/module_importer.js
alanhogan: @eoneill I considered that. However, if you look at the test I adde to the suite, I’m passing it in as an array (as expected) and it still totally works. 

There is some magic happening somewhere.

I’m loathe to add extra code paths that may never be needed just because I don’t understand where the magic is happening.

I also think that… if something changes in the future, this line should totally break — now that it's covered by a test, nothing should be hurt by that. And then we can continue on with one code path here.

Sound good?

I can totally remove or update that comment if you want.

lib/module_importer.js
alanhogan: Will do. My JavaScript has a foot in the past! I’ve always been coding for maximum browser compatibility, but this is node — and now Babel is a thing. Gotta rewire my brain :D

alanhogan:
@eoneill Thanks for the review! I’ve pushed commits in response to your feedback :+1:

alanhogan:
@joannajw Want to take a look?

eoneill:
Barring any objections, I'm going to merge this EOD.

eoneill:
Thanks @alanhogan!

alanhogan:
Excellent. Thanks. Maybe we can bump the version number? Then we can `require` it and use a published version :)

Issue #54, opened 2015-08-24
No longer correctly processing subfolders in require statements


mikl:
This used to work, and works with standard Sass.

In my Sass files, I have a `@import "toolkit/kickstart";` statement, which is supposed to load the file from `node_modules/sass-toolkit/stylesheets/toolkit/_kickstart.scss`.

But Eyeglass seems to strip out the `toolkit/` part when searching for the file, as you can see in the error message:

```
Error: Could not import toolkit/kickstart from any of the following locations: /Users/mikl/Work/Liip/Freitag/vkosf6x3zlyoi/repository/themes/neocortex/sass/partials/global/toolkit/kickstart.scss
sass/partials/global/toolkit/kickstart/index.scss
sass/partials/global/toolkit/kickstart.sass
sass/partials/global/toolkit/kickstart/index.sass
sass/partials/global/toolkit/kickstart.css
sass/partials/global/toolkit/kickstart/index.css
sass/partials/global/toolkit/_kickstart.scss
sass/partials/global/toolkit/kickstart/_index.scss
sass/partials/global/toolkit/_kickstart.sass
sass/partials/global/toolkit/kickstart/_index.sass
sass/partials/global/toolkit/_kickstart.css
sass/partials/global/toolkit/kickstart/_index.css
node_modules/sass-toolkit/stylesheets/kickstart/index.scss
node_modules/sass-toolkit/stylesheets/kickstart.scss
node_modules/sass-toolkit/stylesheets/kickstart/index.sass
node_modules/sass-toolkit/stylesheets/kickstart.sass
node_modules/sass-toolkit/stylesheets/kickstart/index.css
node_modules/sass-toolkit/stylesheets/kickstart.css
node_modules/sass-toolkit/stylesheets/kickstart/_index.scss
node_modules/sass-toolkit/stylesheets/_kickstart.scss
node_modules/sass-toolkit/stylesheets/kickstart/_index.sass
node_modules/sass-toolkit/stylesheets/_kickstart.sass
node_modules/sass-toolkit/stylesheets/kickstart/_index.css
node_modules/sass-toolkit/stylesheets/_kickstart.css
```

If Eyeglass had kept the `toolkit/` part from the import statement, this would work fine.

eoneill:
This looks like it's behaving correctly to me.

[...]

PR #56, opened 2015-08-25
Make importer work with @import "foo/bar"

alanhogan:
In particular when used in conjunction with includePaths.

Includes a test.

### DIFF
--- a/lib/util/files.js
+++ b/lib/util/files.js
@@ -35,7 +35,11 @@ function namify(cb) {
  **/
 function expandFileName(uri, origin, location) {
   var names = [];
+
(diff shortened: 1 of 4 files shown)


alanhogan:
@chriseppstein This solved my problem, but I am getting the increasing suspiction that my original fix #51 solved a problem in a suboptimal way, because the module importer code and `utils/files` seem written with assumptions around *eyeglass modules* more so than generic *sass importing*. I think the diff in this PR will make it clear what I mean.

Incidentally, I believe this PR would solve the problem in #54, although perhaps not in the ideal way? Your feedback is welcome

PR #57, opened 2015-08-25
Allow importing when there is a dot in a folder name

alanhogan:
This change may yield some number of somewhat nonsensical searched filenames in some cases (like `.scss.scss`) but it also allows imports like `my.folder/my.styles` I believe it may also help with Asset Pipeline-influenced names like `.css.scss`.

### DIFF
--- a/lib/util/files.js
+++ b/lib/util/files.js
@@ -1,6 +1,7 @@
 "use strict";
 
 var path = require("path");
+var lodash = require("lodash");
 
 function namify(cb) {
   ["", "_"].forEach(function(prefix) {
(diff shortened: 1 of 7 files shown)


alanhogan:
Note this PR is a strict superset of #56

alanhogan:
@chriseppstein I ran into a lot of file-not-found sort of errors when moving a big project from ruby sass / compass. All these changes helped!

Nasty patterns that were essentially out of my control, Sass-wise:

- Styles shared via `bower_components` with dots in the bower module name (`shared.styles/font.weights`)
- Sass files named in an Asset Pipeline-influenced way: `shared-styles.css.scss`

In short, this PR helps find more of the files that might otherwise go unfound. And it doesn’t break any existing tests :)

PR #58, opened 2015-08-26
Fixed linting errors, added an info dump of the node-sass library in gulp

Jakobo:
Fixed linting errors, added an info dump of the node-sass library in gulp

Frequently as part of testing upgrades to node-sass / libsass, we will want to know
what version of the libraries we are running against. This change to the gulpfile
will print out the `.info` property of the node-sass library.

Additionally, I fixed all our linting errors.


eoneill:
Thanks a bunch for doing this!

PR #59, opened 2015-08-28
Removing circular reference to options when sassOptions is called

### LINKED ISSUE #52: Circular reference in sassOptions breaks Grunt
In this commit, a reference to the Eyeglass object is made inside the options object https://github.com/sass-eyeglass/eyeglass/commit/9479bbfb45ae70b07be3b77f884ce588491800ae#diff-6d186b954a58d5bb740f73d84fe39073R31

Because the options object is also assigned to the eyeglass object, this creates a circular reference. So if you want to walk the options object, you would get an infinite recursion in the form of `eyeglass.options.eyeglass.options.eyeglass.options.eyeglass.options` etc.

Grunt, being the busybody it is, wants to walk the options, resulting in this error message:

> Fatal error: Circular reference detected (.eyeglass.options)

If nothing else, the circular reference should be removed when calling `.sassOptions()`.

JohnRiv:
This fixes #52

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -40,7 +40,11 @@ Eyeglass.prototype = {
   },
   sassOptions: function() {
     // TODO remove eyeglass specific options or maybe namespace them?
-    return this.options;
+    /*eslint-disable */
+    var options = require('util')._extend({}, this.options);
+    /*eslint-enable */
+    delete options.eyeglass;
+    return options;
   },
   normalizeOptions: function(options) {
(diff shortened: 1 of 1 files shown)


chriseppstein:
Thanks for the patch, I opted for a different solution. Please let me know whether it works for you.

22d0c3fd3ab44ac11579f435da68a03355824230

JohnRiv:
I upgraded to 0.6.1 and that fixed the circular reference issue, thank you

PR #60, opened 2015-09-10
Handling the various different states of includePaths

JohnRiv:
Pull Request #51 to solve issue #47 is causing issues in Grunt for me. Specifically, if I include includePaths like this:

```
  var eyeglass = new Eyeglass({
    includePaths: [path.join(rootDir, 'node_modules','compass-mixins','lib')]
  });
```

I get a `Fatal error: undefined is not a function`

If I exclude includePaths like this:

```
  var eyeglass = new Eyeglass({});
```
I get a `Fatal error: Cannot read property 'forEach' of undefined`

This pull request maintains the logic in #51 when includePaths is a String (which is in the case in the tests), but also handles the previous logic of includePaths being an Array (which is the case when I run it with Grunt).

For reference, here are the versions I'm running:
eyeglass 0.6.1
grunt-sass 1.0.0
node-sass 3.3.2
libsass 3.2.5

@alanhogan @FransTwisk, if you can check if this works with your configurations, that would help.

### DIFF
--- a/lib/module_importer.js
+++ b/lib/module_importer.js
@@ -63,8 +63,9 @@ function makeImporter(eyeglass, sass, options, fallbackImporter) {
     var relativePath = fragments.slice(1).join("/");
     var pkgRootDir = isRealFile ? importUtils.packageRootDir(path.dirname(prev)) : root;
     var jsFile = importUtils.getModuleByName(moduleName, pkgRootDir);
-    // includePaths here is a string, despite the config option being an Array
-    var includePaths = options.includePaths.split(path.delimiter);
+    // in some cases, includePaths here is a string, despite the config option being an Array
+    // other cases, it is an Array (or undefined if it was not passed as an option)
+    var includePaths = (options.includePaths && options.includePaths.split) ? options.includePaths.split(path.delimiter) : options.includePaths || [];;
 
     if (jsFile) {
       // is eyeglass module
--- a/test/fixtures/app_assets/sass/no_includePaths.scss
+++ b/test/fixtures/app_assets/sass/no_includePaths.scss
@@ -1,0 +1,5 @@
+$c: red;
+
+.foo {
+  color: $c;
+}
--- a/test/test_imports.js
+++ b/test/test_imports.js
@@ -16,6 +16,18 @@ describe("core api", function () {
    testutils.assertCompiles(options, expected, done);
  });
 
+ it("should compile a sass file if includePaths was not passed as an option & not needed", function(done) {
+   var expected = ".foo {\n" +
+                  "  color: red; }\n";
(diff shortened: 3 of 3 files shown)


lib/module_importer.js
```diff
@@ -63,8 +63,9 @@ function makeImporter(eyeglass, sass, options, fallbackImporter) {
     var relativePath = fragments.slice(1).join("/");
     var pkgRootDir = isRealFile ? importUtils.packageRootDir(path.dirname(prev)) : root;
     var jsFile = importUtils.getModuleByName(moduleName, pkgRootDir);
-    // includePaths here is a string, despite the config option being an Array
-    var includePaths = options.includePaths.split(path.delimiter);
+    // in some cases, includePaths here is a string, despite the config option being an Array
+    // other cases, it is an Array (or undefined if it was not passed as an option)
+    var includePaths = (options.includePaths && options.includePaths.split) ? options.includePaths.split(path.delimiter) : options.includePaths || [];;
```
eoneill: Thanks for fixing this.

One suggestion: can we move this logic into a function:

```js
function normalizeIncludePaths(paths) {
  paths = paths || [];

  // if it's string like...
  if (paths.split) {
    // split it into an array
    paths = paths.split(paths.delimiter);
  }
  
  // ensure that we're always using an array
  if (!Array.isArray(paths)) {
    console.warn("includePaths should be an array but got " + typeof paths);
    return [];
  }

  return paths;
}

...

var includePaths = normalizeIncludePaths(options.includePaths);
```

it's a bit more verbose but there's a clear abstraction for doing this normalization

alanhogan: I’m still confused on why this is sometimes a string and sometimes an array. That seems like a core … wrongness… and by cleaning that up, we can write more concise and reliable code here.

alanhogan:
Sorry my PR broke your stuff. It’s confusing to me how includePaths specified as an array ended up as a string for me, but not for you. I’d like to see that inconsistency resolved as the true fix here, but have no objection to this as an immediate fix!

JohnRiv:
no worries, it's not clear to me why its format is different in different cases either. I'll update the code tonight as suggested.

chriseppstein:
I beat you to it @JohnRiv :) See commit: 010ea9452888a936d9c75db8b800417f7548df08

This is now merged, I just made my iterations on top of your patches.

lib/module_importer.js
chriseppstein: The [node-sass documentation](https://github.com/sass/node-sass#includepaths) says it should be an Array<String>. But handling the SASS_PATH format is also easy enough so I just handle both, but normalize it all to an array at instantiation time.

JohnRiv:
Ha thanks Chris, that worked :+1:

PR #67, opened 2015-10-07
Filesystem Module

chriseppstein:
This module for eyeglass implements the basics of the filesystem API I'd like to add to eyeglass. This forms the foundation of what is needed for a eyeglass module to implement an assets override in pure sass (the assets code allows an eyeglass module to create an assets.scss file that overrides the default assets importer). When designing the assets module, we planned to make it easy for developers to use Sass to maintain their assets API and issue deprecation warnings, etc, all without dropping into javascript.

Beyond assets, there's obviously lots of things that I think people will use this for. Especially once we add more functionality to it (E.g. directory listings).

But I wanted to get initial feedback to make sure no one hates what I've made so far.

There's a bunch of docs in the pull request itself.

### DIFF
--- a/lib/functions/fs.js
+++ b/lib/functions/fs.js
@@ -2,6 +2,7 @@
 
 var fs = require("fs");
 var path = require("path");
+var glob = require("glob");
 
 // Returns whether a file exists.
 function existsSync(file) {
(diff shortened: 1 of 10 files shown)


lib/functions/fs.js
```diff
@@ -0,0 +1,179 @@
...
+module.exports = function(eyeglass, sass) {
+  var sassUtils = require("node-sass-utils")(sass);
+
+  function inSandbox(fsPath) {
+    var sandbox = eyeglass.options.fsSandbox;
+    if (!sandbox) {
+      return true;
+    }
+    if (Array.isArray(sandbox)) {
+      for (var i = 0; i < sandbox.length; i ++) {
```
eoneill: can we use `.some()` here instead?

```js
return sandbox.some(function(sb) {
  if (pathInSandboxDir(fsPath, sb)) {
    return true;
  }
});
```

lib/functions/fs.js
```diff
@@ -0,0 +1,179 @@
...
+          segments.unshift(expandedPath);
+          var resolved = path.resolve.apply(path, segments);
+          done(sass.types.String(resolved));
+        } else {
+          done(new Error("No path is registered for " + pathId));
+        }
+      },
+    "eyeglass-fs-join($segments...)": function(segments, done) {
+      var jsSegments = sassUtils.castToJs(segments);
+      var joined = path.join.apply(path, jsSegments);
```
eoneill: consider passing `null` for `this` context...

```js
path.join.apply(null, jsSegments)
```

lib/functions/fs.js
```diff
@@ -0,0 +1,179 @@
...
+      if (inSandbox(filename)) {
+        fs.readFile(filename, function(err, contents) {
+          if (err) {
+            done(sass.types.Error(err.message));
+          } else {
+            done(sassUtils.castToSass(contents.toString()));
+          }
+        });
+      } else {
+        done(sass.types.Error("Security violation: Cannot access " + filename));
```
eoneill: can we DRY up this error as we use it several times?

e.g.
```js
function accessViolation(location) {
  return sass.types.Error("Security violation: Cannot access " + location));
}
```

eoneill:
Left a few minor comments but looks good to me.

Issue #70, opened 2015-10-15
Some imports still fail unexpectedly


alanhogan:
We've got a bower-managed package named `zd-css-settings.weights` containing a SCSS file we want to import. 

The line is: 
`@import '../../../bower_components/zd-css-settings.weights/settings.weights';`

and it blows up thusly:

```
module.js:340
    throw err;
    ^
Error: Cannot find module 'zd-css-settings.weights/package.json'
    at Function.Module._resolveFilename (module.js:338:15)
    at resolve (/harmony/code/zendesk_console/node_modules/broccoli-eyeglass/node_modules/eyeglass/lib/util/resolve.js:19:17)
    at /harmony/code/zendesk_console/node_modules/broccoli-eyeglass/node_modules/eyeglass/lib/util/discover.js:151:17
    at Array.forEach (native)
    at discover (/harmony/code/zendesk_console/node_modules/broccoli-eyeglass/node_modules/eyeglass/lib/util/discover.js:150:25)
    at Object.getModules [as all] (/harmony/code/zendesk_console/node_modules/broccoli-eyeglass/node_modules/eyeglass/lib/util/discover.js:198:19)
    at Object.ImportUtilities.getModuleByName (/harmony/code/zendesk_console/node_modules/broccoli-eyeglass/node_modules/eyeglass/lib/import_utils.js:38:29)
    at Object.<anonymous> (/harmony/code/zendesk_console/node_modules/broccoli-eyeglass/node_modules/eyeglass/lib/module_importer.js:66:30)
    at options.importer (/harmony/code/zendesk_console/node_modules/node-sass/lib/index.js:321:31)
```

alanhogan:
Having a hard time reproducing in the test suite though.

eoneill:
Does `zd-css-settings.weights` have an `eyeglass-exports.js`?

I'd like to understand what's going on here more, but you might be able to work around the issue by adding `../../../bower_components` to your `includePaths` (might want to use an absolute path).

Then here you'd `@import "zd-css-settings.weights/settings.weights";`

eoneill:
Additionally, is `zd-css-settings.weights` in your `package.json`?

Looking through the logic, it should only be trying to resolve `zd-css-settings.weights/package.json` if `zd-css-settings.weights` was declared in your `dependencies` or `devDependencies`.

alanhogan:
Thanks for your responses, @eoneill —

[...]

Issue #66, opened 2015-09-30
Add a way to list discovered modules for debugging purposes


kohlmannj:
[Per my lament on Twitter](https://twitter.com/jkohlmann/status/649305828915343364), it'd be great if the Eyeglass instance had an attribute or method to **list the Eyeglass modules it knows about** (and some other information, like their file paths).

kohlmannj:
Added thought: debug output for known functions and Sass paths per named Eyeglass module would be good, something like:

```javascript
"modules": {
    "eyeglass-sample": {
        "version": "0.0.3",
        "sasspath": "node_modules/eyeglass_example/sass",
        "functions": [
            "hello($name)"
        ]
    },
    "susy": {
        "version": "2.2.6",
        "sasspath": "node_modules/susy/sass"
    }
```

kohlmannj:
Also adding what I noted on Twitter: the inspiration here was that the libsass compiler's import error, which lists the load paths it tried, didn't include any of the paths to Eyeglass modules. More / original tweets here: https://twitter.com/jkohlmann/status/649359893682606080

eoneill:
eyeglass _should_ output an error when it can't find a module / import path. This error will look something like...

```sh
Error: src/scss/main.scss
  1:9  Could not import foo from any of the following locations:
  /Users/eoneill/workspace/eyeglass-test/src/scss/foo
  /Users/eoneill/workspace/eyeglass-test/src/scss/foo.scss
  /Users/eoneill/workspace/eyeglass-test/src/scss/foo/index.scss
  /Users/eoneill/workspace/eyeglass-test/src/scss/foo.sass
  /Users/eoneill/workspace/eyeglass-test/src/scss/foo/index.sass
  /Users/eoneill/workspace/eyeglass-test/src/scss/foo.css
  /Users/eoneill/workspace/eyeglass-test/src/scss/foo/index.css
  /Users/eoneill/workspace/eyeglass-test/src/scss/_foo.scss
  /Users/eoneill/workspace/eyeglass-test/src/scss/foo/_index.scss
  /Users/eoneill/workspace/eyeglass-test/src/scss/_foo.sass
  /Users/eoneill/workspace/eyeglass-test/src/scss/foo/_index.sass
  /Users/eoneill/workspace/eyeglass-test/src/scss/_foo.css
  /Users/eoneill/workspace/eyeglass-test/src/scss/foo/_index.css
  ...
```
[...]

PR #72, opened 2015-10-21
[WIP] refactor module discovery

eoneill:
This is a pretty substantial change, but it should address some weird edge cases we've seen along the way with module discovery (e.g. #70).

The current approach attempts to discover modules on every `@import` statement. This is both costly and non-deterministic.

The approach proposed here is to discover modules once (when `new Eyeglass()` instance is created). This collection of modules is the source of truth for the life-span of the instance (stored on `this.modules`). We use this for all module lookups and we preserve the tree hierarchy to determine scope of a module.

This also allows us to expose a tree view of the modules for easy debugging (per #66). For example, here's the result of `this.modules.getGraph()` on the `function_modules` test case:
```
:root(function_modules)
├─┬ fn_module_a
│ └── transitive_fn
├── fn_module_b
└── eyeglass
```

Please provide any feedback and I'll work on polishing this up a bit and ensuring all the test are passing again.


lib/util/eyeglass_modules.js
```diff
@@ -0,0 +1,283 @@
...
+var resolve = require("./resolve");
+var packageUtils = require("./package");
+var path = require("path");
+var merge = require("lodash.merge");
+var includes = require("lodash.includes");
+var semver = require("semver");
+var uuid = require("uuid");
+
+var EYEGLASS_KEYWORD = "eyeglass-module";
+var ROOT_ID = ":root-" + uuid.v4();
```
Jakobo: reasons for uuid v reserved name or just non-npm compatible package name?

lib/util/eyeglass_modules.js
```diff
@@ -0,0 +1,283 @@
...
+        return flattenModules(dependencies[name], registry);
+      }, registry || {});
+      return result;
+    }
+  }
+
+  return registry;
+}
+
+function checkVersions(versions, latest) {
```
Jakobo: `checkVersions` seems to return incompatible versions. Either the fn name or the return object should probably change.

eoneill:
Part of me wants to move `EyeglassModules` into a separate repo with a dedicated test suite (e.g. `eyeglass-core-modules`)

lib/util/eyeglass_modules.js
```diff
@@ -0,0 +1,283 @@
...
+      return {
+        name: mod.name,
+        left: mod,
+        right: latest
+      };
+    }
+  });
+}
+
+function dedupeModules(modules) {
```
Jakobo: This operates in place and returns an error array. Will definitely want to make sure that's noted and has a test

lib/util/eyeglass_modules.js
```diff
@@ -0,0 +1,283 @@
...
+};
+
+// this is mostly for debugging purposes
+EyeglassModules.prototype.getGraph = function() {
+  var hierarchy = getHierarchy(this.tree);
+  hierarchy.label = getDecoratedRootName.call(this);
+  return require("archy")(hierarchy);
+};
+
+function getDecoratedRootName() {
```
Jakobo: Semantics. Is the project moving towards `.call` to create pseudo private scope?

lib/util/eyeglass_modules.js
```diff
@@ -0,0 +1,283 @@
...
+  if (options.isRoot) {
+    // ensure eyeglass itself is a direct dependency
+    dependencies.eyeglass = dependencies.eyeglass || getEyeglassSelf(options);
+  }
+
+
+  return Object.keys(dependencies).length ? dependencies : null;
+}
+
+function tick(str) {
```
Jakobo: Is this the only place we're bothering with backticks around path stuff or are we missing a util file as part of the refactor?

lib/util/package.js
```diff
@@ -0,0 +1,50 @@
+"use strict";
```
Jakobo: 6 months after the initial discovery code was written, this is worthy of 100 <3s. Thank you.

package.json
```diff
@@ -28,25 +28,29 @@
...
     "grunt-release": "^0.13.0",
     "gulp": "^3.9.0",
     "gulp-eslint": "^1.0.0",
-    "gulp-istanbul": "^0.10.0",
+    "gulp-istanbul": "^0.10.1",
     "gulp-mocha": "^2.1.3",
-    "mocha": "^2.2.5",
-    "should": ">= 0.0.1"
+    "mocha": "^2.3.3",
+    "should": ">=7.1.0"
```
Jakobo: Is leaving `>=` intentional? Can we use `^7.1.0` instead?

lib/util/eyeglass_modules.js
eoneill: I was going somewhere with that, but I backtracked and probably don't need it anymore.

lib/util/eyeglass_modules.js
eoneill: Yeah, this definitely needs a test. I expect to be adding several tests as part of this.

lib/util/eyeglass_modules.js
eoneill: I like it, but not entirely opposed to exposing things on the prototype.

lib/util/eyeglass_modules.js
eoneill: I was using it in more places, but then backtracked on the backticks. We should be consistent throughout eyeglass warnings / errors.

package.json
eoneill: I should definitely be using `^7.1.0`. Will update.

chriseppstein:
> Part of me wants to move `EyeglassModules` into a separate repo with a dedicated test suite (e.g. `eyeglass-core-modules`)

I'd rather we didn't.

chriseppstein:
I had set up the travis build so that coverage was only ran once instead of with every build permutation and it looks like you've removed that. :-1:

eoneill:
> I had set up the travis build so that coverage was only ran once instead of with every build permutation and it looks like you've removed that. :-1:

I'll revert this change.

lib/util/eyeglass_modules.js
```diff
@@ -0,0 +1,283 @@
...
+}
+
+function resolveModule(pkg, isRoot) {
+  pkg = packageUtils.getPackage(pkg);
+  var isEyeglassMod = isEyeglassModule(pkg.data);
+  if (isEyeglassMod || (pkg.data && isRoot)) {
+    var modulePath = path.dirname(pkg.path);
+    if (typeof pkg.data.version === "object") {
+      console.log("pkg.data.version", pkg.data.version);
+    }
```
chriseppstein: remove debugging code?

lib/util/eyeglass_modules.js
```diff
@@ -0,0 +1,283 @@
...
+    return {
+      name: getModuleName(pkg) || (isRoot && ROOT_ID),
+      rawName: pkg.data.name,
+      path: modulePath,
+      version: pkg.data.version,
+      main: getModuleExports(pkg.data, modulePath),
+      isEyeglassModule: isEyeglassMod,
```
chriseppstein: why would `isEyeglassModule` ever be false?

eoneill: If the project root itself is not an eyeglass module. That's the only case.

This is important so that we can correctly preserve the dependency tree for checking access permission.

lib/util/eyeglass_modules.js
```diff
@@ -0,0 +1,283 @@
...
+      return true;
+    }
+  });
+}
+
+function flattenModules(branch, registry) {
+  if (branch) {
+    // if the branch itself is a module, add it...
+    if (branch.isEyeglassModule) {
+      registry = registry || {};
```
chriseppstein: better to set this default at the start of the function instead of repeating it several times.

lib/util/eyeglass_modules.js
```diff
@@ -0,0 +1,283 @@
...
+      registry[branch.name] = registry[branch.name] || [];
+      registry[branch.name].push(branch);
+    }
+
+    var dependencies = branch.dependencies;
+
+    if (dependencies) {
+      var result = Object.keys(dependencies).reduce(function(registry, name) {
```
chriseppstein: pretty sure using the same argument name in a nested function is going to be a linter error -- are you running the linter in development?

chriseppstein:
When running the tests they all pass but there's some console logging:

```
{ [Error: Cannot find module 'module_a/package.json'] code: 'MODULE_NOT_FOUND' } [ 'Error: Cannot find module \'module_a/package.json\'',
  '    at Function.Module._resolveFilename (module.js:337:15)',
  '    at resolve (/Users/ceppstei/Projects/eyeglass/lib/util/resolve.js:9:238)',
  '    at /Users/ceppstei/Projects/eyeglass/lib/util/eyeglass_modules.js:9:14880',
  '    at Array.reduce (native)',
  '    at discoverModules (/Users/ceppstei/Projects/eyeglass/lib/util/eyeglass_modules.js:9:14749)',
  '    at resolveModule (/Users/ceppstei/Projects/eyeglass/lib/util/eyeglass_modules.js:9:13389)',
  '    at new EyeglassModules (/Users/ceppstei/Projects/eyeglass/lib/util/eyeglass_modules.js:9:929)',
  '    at Object.Eyeglass (/Users/ceppstei/Projects/eyeglass/lib/index.js:9:1343)',
  '    at decorator (/Users/ceppstei/Projects/eyeglass/lib/index.js:9:4313)',
  '    at Object.module.exports.sassOptions (/Users/ceppstei/Projects/eyeglass/test/testutils.js:9:3084)' ]
```

This occurs on the test `errors on invalid module`. Is this output expected? If so, we should capture stderr/out and assert that it's output. There are test utilities for capturing output if you didn't see them yet.

lib/util/eyeglass_modules.js
eoneill: Yes I am. It doesn't complain, but I was going to change this anyway.

eoneill:
> When running the tests they all pass but there's some console logging:

I need to clean this up still. Tests are not all good yet and I have more tests to write. At this point, I'm mainly looking for preliminary feedback that this is an overall sound approach. There's a lot of cleanup to come, but I don't want to jump down a rabbit hole if the idea is off base.

chriseppstein:
>  I don't want to jump down a rabbit hole if the idea is off base.
 
I don't see any theoretical issues with it. I was memoizing module information already so it's not clear how big of a performance gain this will actually provide, but it does seem to be a nicer abstraction to work with in code.

eoneill:
All tests are now passing with `node-sass@3.3.3`, unfortunately `node-sass@3.4.0` causes tests to fail `@HEAD`, so please run tests locally to verify.

I think I addressed all the comments above, but let me know if I missed anything.

eoneill:
Barring node-sass issues, this should be good for review/merge now. I added code docs as well as some additional tests for `EyeglassModules`

Thanks for your patience.

docs/TROUBLESHOOTING.md
```diff
@@ -8,3 +8,14 @@ This happens during your `npm install` process, and in our case is almost certai
 If you've got a `gyp ERR`, take a look at the [open gyp issues for node-sass](https://github.com/sass/node-sass/search?o=desc&q=gyp&s=updated&state=open&type=Issues&utf8=✓). If you still don't see anything, please create a ticket and both use and the `node-sass` folks will figure out what's up.
 
 If there is no sign of `gyp ERR` anywhere in your install log, then [file an issue](https://github.com/sass-eyeglass/eyeglass/issues) and paste your `npm install` log. It's probably broken for a lot of people. =\
+
+# Logging debug messages
+
+If you're still running into issues, you can gain some insight by enabling some debug flags. To enable this logging, set an environment variable named `DEBUG` to one or more of the following:
```
chriseppstein: Let's link to the [debug module here](https://www.npmjs.com/package/debug).

PR #86, opened 2015-11-16
fix: move debug to dependencies

eoneill:
`debug` was incorrectly added to `devDependencies`, needs to be in `dependencies`


chriseppstein:
Fixed in 991aac0f58443c81441c87fd5b17f7c9cf317999

PR #89, opened 2015-11-16
ensure custom importers are invoked with the correct context. fixes #83, fixes #65

### LINKED ISSUE #65: import-once broken when compiling multiple files per eyeglass instance
Hi,
I've an issue on ***import-once***.

I have two different scss files that will be compiled in two different css files.
In both of these .scss files (for example: bootstrap.scss and application.scss) I need to import a single file called: _variables.scss.

The import-once module seems to prevent this behaviour and in the second file that requires _variables.scss I obtain 
```scss
undefined variable: $variableName
```

is there any workaround for doing this?

Gius

### LINKED ISSUE #83: Pass options to the custom importer
As separate issue from  https://github.com/sass-eyeglass/eyeglass/issues/42#issuecomment-156851307

I try to use this custom importer with eyeglass: https://github.com/at-import/node-sass-import-once

````
[...]
  var eyeglass = new Eyeglass({
    importer: importOnce,
    importOnce: {
      index: true,
      css:   false,
      bower: false
    }
  });
[...]
````

However, the importer isn't able to see the importOnce property in options object, 
is there a way to pass it to the custom importer when eyeglass is used instead of plain node-sass?

````
/home/build/src/test/node_modules/node-sass-import-once/index.js:275
  if (!this.options.importOnce) {
                   ^

TypeError: Cannot read property 'importOnce' of undefined
    at Object.importer [as fallbackImporter] (/home/build/src/test/node_modules/node-sass-import-once/index.js:275:20)
[...]
````

Thank you and
With best regards

### PR DESCRIPTION
**Pull request by eoneill**
This ensures that each custom importer is invoked with the correct node-sass context. This directly fixes #83 and happens to fix #65 as a result of moving the cache onto this context, rather than persisting it within the eyeglass instance.

### CODE REVIEW

File: lib/import_utils.js [hunk 1]
```diff
@@ -2,13 +2,19 @@
...
   this.root = options.root;
   this.eyeglass = eyeglass;
   this.sass = sass;
   this.alreadyImported = {};
   this.fallbackImporter = fallbackImporter;
+  this.context = merge(context, {
+    eyeglass: {
+      imported: {} // keep track of files already imported @see importOnce
+    }
+  });
```
chriseppstein: This merge will happen many times during a compile. won't that cause problems?

chriseppstein: Ah, nvm, I see from the merge documentation that because `imported` is an object this is basically a no-op.

PR #91, opened 2015-11-17
Optional eyeglass exports

chriseppstein:
This PR makes simple eyeglass modules even simpler by not requiring an eyeglass exports file. For those modules that only provide sass files, they can just set a relative path in their package.json file. I suspect this will ease adoption because a small addition to package.json feels a simpler than adding a new javascript file especially to people who don't know js.

### DIFF
--- a/lib/module_importer.js
+++ b/lib/module_importer.js
@@ -65,23 +65,25 @@ function makeImporter(eyeglass, sass, options, fallbackImporter) {
     var jsFile = eyeglass.modules.access(moduleName, isRealFile ? prev : root);
     var sassDir;
 
-    if (jsFile && jsFile.eyeglass.sassDir) {
-      sassDir = jsFile.eyeglass.sassDir;
-    } else if (jsFile && jsFile.main) {
-      var mod = require(jsFile.main)(eyeglass, sass);
-      sassDir = mod.sassDir;
-    }
+    if (jsFile) {
+      if (jsFile.eyeglass.sassDir) {
+        sassDir = jsFile.eyeglass.sassDir;
+      } else if (jsFile.main) {
+        var mod = require(jsFile.main)(eyeglass, sass);
+        sassDir = mod.sassDir;
+      }
 
-    if (jsFile && !(sassDir || isRealFile)) {
-      // No sass directory, give an error
-      importUtils.fallback(uri, prev, done, function() {
-        var missingMessage = "sassDir is not specified in " + jsFile.name + "'s package.json";
-        if (jsFile.main) {
-          missingMessage += " or " + jsFile.main;
-        }
-        done(new Error(missingMessage));
-      });
-      return;
+      if (!sassDir && !isRealFile) {
+        // No sass directory, give an error
+        importUtils.fallback(uri, prev, done, function() {
+          var missingMessage = "sassDir is not specified in " + jsFile.name + "'s package.json";
+          if (jsFile.main) {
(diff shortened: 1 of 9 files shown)


lib/module_importer.js
```diff
@@ -63,18 +63,28 @@ function makeImporter(eyeglass, sass, options, fallbackImporter) {
...
-      if (!sassDir) {
-        importUtils.fallback(uri, prev, done, function() {
-          done(new Error("sassDir is not specified in " + jsFile.main));
-        });
-        return;
-      }
+      sassDir = mod.sassDir;
+    }
+
+    if (jsFile && !(sassDir || isRealFile)) {
```
eoneill: instead of...

```js
if (jsFile && jsFile.eyeglass.sassDir) {
   ...
} else if (jsFile && jsFile.main) { ... }

if (jsFile && !(sassDir || isRealFile)) { ... }
```

consider...

```js
if (jsFile) {
  if(jsFile.eyeglass.sassDir) {
    ...
  } else if (jsFile.main) { ... }
  
  if (!(sassDir || isRealFile)) { ... }
}
```

lib/util/eyeglass_modules.js
```diff
@@ -381,25 +381,37 @@ function dedupeModules(modules) {
   * @param   {Object} pkg - the package.json reference
   * @returns {String} the name of the module
   */
-function getModuleName(pkg) {
+function getModuleName(pkg, pkgPath) {
```
eoneill: `getModuleName` shouldn't need `pkgPath` to complete its job (see comment below)

if it does, you can also reference it off of `pkg.path`

lib/util/eyeglass_modules.js
```diff
@@ -381,25 +381,37 @@ function dedupeModules(modules) {
...
 /**
   * normalizes a given `eyeglass` reference from a package.json
   *
-  * @param   {Object} options - the eyeglass options from the package.json
+  * @param   {Object} options - The eyeglass options from the package.json
+  * @param   {String} pkgPath - The location of the package.json.
   * @returns {Object} the normalized options
   */
-function normalizeEyeglassOptions(options) {
+function normalizeEyeglassOptions(options, pkgPath) {
```
eoneill: I think `pkgPath` should be optional

e.g.

```js
if (pkgPath && normalizedOpts.sassDir) {
  ...
}
```

lib/util/eyeglass_modules.js
```diff
@@ -449,19 +465,19 @@ function getFinalModule(name) {
...
   *
-  * @param   {String} pkg - the path to the modules package.json location
+  * @param   {String} pkgPath - the path to the modules package.json location
   * @param   {Boolean} isRoot - whether or not it's the root of the project
   * @returns {Object} the resolved module definition
   */
-function resolveModule(pkg, isRoot) {
-  pkg = packageUtils.getPackage(pkg);
+function resolveModule(pkgPath, isRoot) {
+  var pkg = packageUtils.getPackage(pkgPath);
```
eoneill: `getPackage` returns an object `{ path: ..., data: ... }`

so you shouldn't need to preserve `pkgPath` here, you can reference `pkg.path`

xzyfer:
Sorry I was on a plane. Out of curiosity has defaulting `sassDir` to the cwd of the `package.json` been discussed? This would be the primary use case for me.

eoneill:
@xzyfer I like that `sassDir` is explicit in this approach.

PR #92, opened 2015-11-18
fix: asset-uri should return the uri not wrapped in url()

eoneill:
`asset-uri` and `asset-url` currently both return `url(...)`

I'd expect `asset-uri` to return only the uri, not wrapped in `url(...)`

This also simplifies `asset-url` to leverage `asset-uri` to get the uri.


chriseppstein:
This was how I had intended it. copypaste error :(

Issue #46, opened 2015-06-29
Simplify API by exporting factory


davidkpiano:
This is a bit of a major change, but I believe that incorporating Eyeglass modules into a project will have less friction by simply exporting a factory function instead of a raw `Eyeglass` object.

**OLD API:**
```js
var sass = require('node-sass');
var Eyeglass = require('eyeglass');
var sassOptions = { ... };
var eyeglass = new Eyeglass(sassOptions); // boilerplate

eyeglass.assets("assets/images", "images", "public/images"); // imperative

// eyeglass.sassOptions() is also boilerplate
sass.render(eyeglass.sassOptions(), function(err, result) { ... });
```

**NEW (proposed) API:**
```js
var sass = require('node-sass');
var eyeglass = require('eyeglass');
var sassOptions = { ... };
var assets = [
  {source: 'assets/images', http: 'images', build: 'public/images'} 
]; // declarative and clear

sass.render(eyeglass(sassOptions, { assets: assets }), function(err, result) { ... });
```

So the arguments would simply be `eyeglass(sassOptions, eyeglassOptions)` and would return a POJO equivalent to `Eyeglass.prototype.options`. Both bits of boilerplate are eliminated, and it also allows the developer to do this shorthand:

```js
var sassOptions = { ... };
var eyeglassOptions = require('eyeglass')(sassOptions);
```

What do you think? I believe it's more intuitive and idiomatic.

Jakobo:
@chriseppstein @eoneill @joannajw

scottdavis:
I dig it.

eoneill:
I like the proposal.

davidkpiano:
@chriseppstein any thoughts on this? I know the `assets` part might be different now but I do believe the API can be simplified.

chriseppstein:
The API did change but I am open to having a way to pass asset declarations in as options because I think this will help grunt integration.

The current API is:

```js
var eyeglass = new Eyeglass({});
eyeglass.assets.addSource("assets", {pattern: "images/**/*"});
```

The `addSource` method passes through to the [AssetPathEntry constructor](https://github.com/sass-eyeglass/eyeglass/blob/master/lib/assets.js#L17-L31).

So it seems like at the highest level, you're asking for a way to specify this via options instead of calling methods. :shrug: ok. We can just iterate over some option and call the method for you.

```js
var sass = require('node-sass');
var eyeglass = require('eyeglass').decorate;
var sassOptions = {
[...]

PR #94, opened 2015-11-20
Simplify API

eoneill:
This is the first pass at simplifying the public API of eyeglass per #46.

# Interface

The exported interface supports two models. As an instance or as a method.

## `new Eyeglass(options)`

Using the `new` keyword we invoke the constructor and return an eyeglass instance. This will allow more complex interactions with the eyeglass instance before passing it off to `node-sass`. The wrapped options are exposed on the instance property `#options`
```js
var Eyeglass = require("eyeglass");
var eyeglass = new Eyeglass(options);
eyeglass.assets.addSource(...);

sass.render(eyeglass.options, function(err, result) { ... });
```

## `eyeglass(options)`

As a method, this will provide the simplest interface for users. This simply wraps the options and returns them.

```js
var eyeglass = require("eyeglass");

sass.render(eyeglass(options), function(err, result) { ... });
```

## `options`

The `options` argument contains the [sass options](https://github.com/sass/node-sass#options) with an `eyeglass` namespace for all the eyeglass options:

```js
{
  /* sassOptions */
  ...
  
  eyeglass: {
    /* eyeglassOptions */
  }
}
```
### example

```js
var eyeglass = new Eyeglass({
  /* sassOptions */
  file: "foo.scss",
  includePaths: ["/some/path"],
  importer: function() { ... },

  eyeglass: {
    /* eyeglassOptions */
    root: __dirname,
    cacheDir: ".my-cache",
    buildDir: "build",
    httpRoot: "/",
    assetsHttpPrefix: "assets",
    assetsRelativeTo: "assets/foo",
    strictModuleVersions: true,
    enableImportOnce: false, // disable import-once
    engines: {
      sass: require("node-sass")
    }
  }

});
```

# back-compat

Currently, this patch introduces the new API, but preserves back-compatibility with deprecation warnings. These deprecation warnings are currently disabled in testing to demonstrate this did not break the supported integration.

I'd like us to remove all the deprecations and back-compat support when we go 1.0. The compat support is pretty cleanly abstracted and should be easy to remove.

## removed methods

This preserves back-compat for the primary methods people are currently using. This removes a bunch of things currently exposed on `Eyeglass.prototype` which I believe should _not_ be public. This includes:

* `#defaultRoot`
* `#defaultCacheDir`
* `#sassPathFromEnv`
* `#defaultIncludePaths`
* `#root`
* `#normalizeOptions`
* `#sassOptions` (currently supported with deprecation warning)

I can add deprecation methods for these if we think it's needed.

# you changed a test?!

Yeah, I modified the test for circular options. I currently don't expose `eyeglass` onto the sass options. We don't seem to use it for anything, so I chose not to expose it. If there's a reason we need to expose it, please comment. If not, I'd like to leave it out until we do have a need to expose it.

# TODO

This PR is not "complete". I wanted to try to reduce the scope of changes introduced here and some other improvements are still pending. These will likely come in a subsequent PR to reduce the overall complexity of the changes.

## add `assets` option

In a subsequent PR, I will add support for passing `assets` via the eyeglassOptions. This will likely look something like...
```js
var eyeglass = new Eyeglass({
  /* sassOptions */

  eyeglass: {
    /* sassOptions */
    root: __dirname,
    ...,
    /* eyeglass assets */
    assets: [
      {
        directory: "assets",
        pattern: "images/**/*.png"
      }
      {
        directory: "public",
        pattern: "fonts/**/*"
      }
    ]
  }
});
```

## clean up tests

The tests are mostly untouched for sanity. In a subsequent PR, I will rewrite the tests to use the new interface and move the options to the `eyeglass` namespace.


scottdavis:
So the first thing i see with out looking at code is that you say you are removing the `#root` option but I'm seeing you use it as an example in more then one place? Im not sure what the intent is.

davidkpiano:
Simplified API looks great. Thanks for this!

eoneill:
Thinking about this again, I'm now inclined to merge `sassOptions` and `eyeglassOptions`, but under an `eyeglass` namespace:

```
var options = eyeglass({
  // sassOptions
  file: "foo.scss",
  includePaths: [...],

  // eyeglassOptions
  eyeglass: {
    root: __dirname,
    cacheDir: ".my-cache",
    engines: {
      sass: require("node-sass")
    },
    ...
  }
});
```

This will likely make build integrations easier, but still maintain the clear distinction between `sassOptions` and `eyeglassOptions`.

Thoughts?

lib/options.js
```diff
@@ -0,0 +1,136 @@
...
+  // if it looks like node-sass...
+  if (arg && arg.render && arg.renderSync && arg.info) {
+    // throw a deprecation warning
+    deprecate([
+      "You should no longer pass `sass` directly to Eyeglass. Instead pass it as an option:",
+      "var options = eyeglass(sassOptions, {",
+      "  engines: {",
+      "    sass: require('node-sass')",
+      "  }",
+      "});"
```
chriseppstein: This warning still has the options passed as separate options.

Also, not sure what the benefit is of passing this as an option, especially in the `engines` namespace.

lib/options.js
```diff
@@ -0,0 +1,136 @@
...
+    "assetsRelativeTo",
+    "strictModuleVersions"
+  ].forEach(function(option) {
+    if (eyeglassOptions[option] === undefined && sassOptions[option] !== undefined) {
+      deprecate([
+        "`" + option + "` should be passed into the eyeglass options rather than the sass options:",
+        "var options = eyeglass(\/* sassOptions *\/, {",
+        "  " + option + ": ...",
+        "});"
+      ].join("\n  "));
```
chriseppstein: this warning still has eyeglass options passed separately.

lib/options.js
```diff
@@ -0,0 +1,136 @@
...
+
+function defaultEyeglassOptions(options, eyeglass) {
+  // default root dir
+  options.root = defaultValue(options.root, process.cwd());
+  // default cache dir
+  options.cacheDir = defaultValue(options.cacheDir, path.join(options.root, ".eyeglass_cache"));
+  // default engines
+  options.engines = defaultValue(options.engines, {});
+  options.engines.sass = defaultValue(options.engines.sass, require("node-sass"));
+  options.engines.eyeglass = defaultValue(options.engines.eyeglass, eyeglass);
```
chriseppstein: yo dawg, optception!

I thought we decided eyeglass didn't need to be in the options.

eoneill:
I went ahead and updated the PR with the eyeglassOptions merged back into sassOptions, under an `eyeglass` namespace

lib/options.js
chriseppstein: Also, this will result in a circular reference to options which is bad.

eoneill: the eyeglass instance isn't on the options, but the eyeglass options are exposed on the options

eoneill: actually, I am exposing the instance, which I shouldn't be. Updated latest to set the `Eyeglass` engine onto the function

not terribly useful, but if someone were to return you an arbitrary instance, you could know which `Eyeglass` was used to construct it

test/testutils.js
```diff
@@ -1,5 +1,7 @@
 "use strict";
 
+// disable eyeglass deprecations until tests are rewritten without deprecated methods
+process.env.EYEGLASS_DEPRECATIONS = "false";
```
chriseppstein: Why not pass this as an option?

lib/options.js
eoneill: Ok, removed it entirely as it's not useful currently. #overegineered

chriseppstein:
I'd like to add specific tests around deprecated APIs to make sure they are working. (deprecation warnings only silenced in that specific test) and migrate all tests to the preferred API.

test/testutils.js
eoneill: you mean something like...

```js
new Eyeglass({
  eyeglass: {
    disableDeprecations: true
  }
});
```

eoneill:
> I'd like to add specific tests around deprecated APIs to make sure they are working. (deprecation warnings only silenced in that specific test) and migrate all tests to the preferred API.

By virtue of our current tests passing, we should be able to verify that all current options and `#sassOptions()` method are working correctly.

My preference would be to get this in as a "nothing has changed" (and the tests certify that), and then do the test cleanup in a second PR.

If we follow that approach, I will rewrite the existing tests to use the new API, and add new tests that leverage the old API and explicitly check that we log appropriate deprecation warnings. At which point, there will be no need to globally disable the warnings.

test/testutils.js
eoneill: The reason I didn't do this currently is that it potentially complicates the options normalization phase. I'd have to normalize the options before I can check `disableDeprecations`, which would mean I'd have to queue up any deprecations encountered up to that point.

This would also mean I'd have to touch the tests in a more significant way, per my comment below.

chriseppstein:
:+1: on a follow-on PR.

test/testutils.js
chriseppstein: This is a public API, right? I'm ok with having a default for it as set in the environment, but it should also have a corresponding option. It's ok to normalize this option first.

eoneill: Are you ok with doing the option in a follow up PR? And I'll remove the environment variable in favor of it.

eoneill: I went ahead and added this support to the current PR with a test case.

chriseppstein:
I made some changes to the deprecation system. deprecation warnings are now issued with a startVersion and a removeVersion. This is logged and the env variable and the option both take a string instead of a boolean which causes warnings for that version and below to be silenced. This way, deprecation warnings in new releases will be noticed.

davidkpiano:
Is there any reason to export the raw class instead of a factory?

chriseppstein:
@davidkpiano the function we export acts as a class when used with `new` and as an options pass-through when called without `new`. What API were you wishing for?

davidkpiano:
Oh, sorry, I missed that. Ignore my comment :smile:

chriseppstein:
It's a clever approach and kind of easy to miss from scanning the code. The docs should make it very clear tho.

PR #102, opened 2015-11-24
Clean up tests to use new API

eoneill:
Per #94, this updates the test cases to use the new simplified API.

This also adds tests for deprecation warnings and tests for back-compat API support.


chriseppstein:
Looks great!

Issue #96, opened 2015-11-21
Explicit modules


davidkpiano:
Hey @chriseppstein  -

In writing [Sassport](https://github.com/davidkpiano/sassport), probably the biggest difference between that project and Eyeglass was the ability to explicitly include modules, which is modeled after what PostCSS does (in terms of including modules). 

The external API of Eyeglass is quite a bit different than Sassport in that Eyeglass is _passed in_ as options in Sass whereas Sassport _wraps_ Sass, but we can still make this work:

```js
var sass = require('node-sass');
var eyeglass = require('eyeglass'); // pending https://github.com/sass-eyeglass/eyeglass/pull/94

// Local modules - part of project, no need/desire to include them via NPM
var fooModule = require('./path/to/foo-eyeglass-module');
var barModule = require('./path/to/bar-eyeglass-module');

var eyeglassOptions = eyeglass(sassOptions, eyeglassOptions)
  .modules([ fooModule, barModule ]);

sass(eyeglassOptions).render(...);
```

Then these modules can be imported in any Sass file just like any other Eyeglass module, via `@import 'foo'`, etc. I think this would be a great additional feature because:

1. it bypasses potentially unnecessary searching through `node_modules`,
2. it allows the module to come from anywhere, whether it's NPM, bower, in-project, etc. etc.
3. No need to add anything extra to a module's `package.json` file
4. NPM isn't really for assets, such as fonts or images. This makes using Eyeglass as it currently is a little out-of-place for assets. 

Would you consider explicit modules?

eoneill:
@scottdavis had a similar ask to do something like this.

I think there's value in it, but we make a lot of assumptions with the current module discovery.

Thinking about this last night, I think the API would look something like...

```js
var Eyeglass = require("eyeglass");
var eyeglass = new Eyeglass(...);

eyeglass.modules.add(function(eyeglass, sass) {
  return {
    name: "my-module",
    functions: { ... },
    assets: [],
    sassDir: ...
  };
});
```
There are intricacies to consider here. I'll reply with a follow up (probably on Monday) with some concerns and how we might handle them.

davidkpiano:
Not the biggest fan of the imperative syntax that implies only one module can be added at a time, though. But yeah, that would be great if it were considered.

eoneill:
`#modules.add` could take an array. I'm more interested in what is actually be passed to it (e.g. a function) and what that entails.

[...]

PR #108, opened 2015-11-26
add support for including modules manually

eoneill:
Per #96, this would support adding modules manually via JavaScript (not discovered through `package.json`).

I'm still not entirely convinced this is a good idea, but it's easy enough to support I'm not too concerned.

The interface is to pass the modules (objects) via the eyeglass options:

```js
var eyeglass = new Eyeglass({
  eyeglass: {
    modules: [
      {
        name: "my-module",
        main: function(eyeglass, sass) {
          return {
            sassDir: ...,
            functions: ....,
            assets: ....
          };
        }
      },
      ...
    ]
  }
});
```


chriseppstein:
Looks good, we need to add docs for this.

eoneill:
Latest update improves `EyeglassModule` constructor to do the heavy lifting.

```js
var mod = new EyeglassModule({ ... });
mod.init(eyeglass, sass);
```

This will normalize the module interface and the `init()` will execute the `main` (if present). This is useful so we don't have to `if (mod.main) { require(mod.main)(eyeglass, sass); }` in multiple places and makes the interface more consistent overall.

This updated interface now also allows us to manually add a module with a path:

```js
var eyeglass = new Eyeglass({
   ...,
   eyeglass: {
    modules: [{
      path: path.join(__dirname, "bower_components/my-bower-module")
    }]
  }
});
```

Finally, this will allow modules to expose custom JavaScript interfaces. e.g.
```js
// module-foo/eyeglass-exports.js
module.exports = function(eyeglass, sass) {
   return {
    ...
    doSomething: function() { ... }
  }
};
```
```js
var eyeglass = new Eyeglass(...);
eyeglass.modules.find("module-foo").doSomething(...);
```

chriseppstein:
@eoneill what is the use case(s) you have in mind for the "custom JavaScript interfaces"

eoneill:
http://www.restylecss.com/documentation/understanding-grammar/

Currently, in reSTYLE, I abuse the `options` object to expose an `addGrammarEngine` method.

With this new approach, I would expose the method on the `eyeglass-exports`:

```js
module.exports = function(...) {
  return {
    ...
    addGrammarEngine: function(...) {}
  }
};
```
And consumers would access this via...
```js
eyeglass.modules.find("restyle").addGrammarEngine(...)
```

chriseppstein:
See my comments on #96. This exposes a new public API of things that used to be only internal. I don't want to do this just because it's "easy" -- there has to be a compelling use case and I'm still not convinced of one. I want to hold back on merging this -- I don't know how to review a feature without having the context of the goals clear in my mind.

chriseppstein:
Closing this, after #112 landed, any features from this PR need to be reworked and resubmitted. I'm also still not sold on the idea of manual modules.

PR #112, opened 2015-12-16
add EyeglassModule constructor for module instances

eoneill:
(this is a rework of some changes proposed in #108)

This doesn't introduce any new functionality, but simplifies what an internal definition of an eyeglass module is.

This introduces a new constructor (`EyeglassModule`) that will return an instance of a module. This allows us to get a consistent module instance and should enable us to better reason about what a module is going forward.


chriseppstein:
This is great. Thanks!

Issue #114, opened 2015-12-29
Eyeglass causes Sass task to fail when run on a directory of files


iammikecohen:
I have followed the guide for adding Eyeglass to a Grunt build process and have eyeglass working fine on any one individual file. When I run my Sass task on a directory (50+ files) the first 11 compile but then the task hangs and no other files are built. There is no error message when the task fails. This might not be enough information to reproduce or diagnose the problem, if that is the case please let me know how I can provide better information.

iammikecohen:
The number of files turned out to be a red herring. I am working on finding the specific issue. It seems to be related to the @import directive. Right now I can build all of my legacy files that don't use partials but a maximum of three SCSS files that use @import

iammikecohen:
After trying different combinations of files all I can say is that when I try to build a directory of SCSS files with eyeglass the Sass task freezes after it encounters the fourth file that imports partials. I have used different combinations of the problematic files and as long as there are no more than three files that use `@import` the build succeeds. All Eyeglass managed modules are imported using relative paths to the bower folder. They will build with node sass but not once eyeglass has been added.

eoneill:
Can you provide a git repo to reproduce this issue? It'll make it easier for us to debug this issue.

chriseppstein:
@iammikecohen ping?

iammikecohen:
This is tied into a monolith of code. I'm going to try to replicate in a new repo over the weekend and share with you by Monday. Apologies for the delayed response.

chriseppstein:
@iammikecohen Thank you! btw, if this is a bug it's probably a node-sass or libsass bug but I'm happy to help get it fixed by them. I see this in their issue tracker: https://github.com/sass/node-sass/issues/857

eoneill:
@iammikecohen can you give this patch a try and see if it addresses the issue: https://github.com/sass-eyeglass/eyeglass/pull/116

This is one of the most bizarre issues I've ever seen :)

chriseppstein:
This should be fixed in release `0.8.1`.

PR #116, opened 2016-01-08
fix issue with fs.readFile hanging (issue #114)

eoneill:
This is a pretty bizarre issue, but it's reproducible with the included test.

When 4 or more files with `@imports` are compiled in parallel, `fs.readFile` blows up and hangs.

This fixes the issue by using `fs.readFileSync` in an async wrapper. There's probably a more elegant way to fix this, but I'm not having any luck right now.


micahgodbolt:
:+1: this fixes the issue for me

chriseppstein:
Fixed in release `0.8.1`.

PR #117, opened 2016-01-08
Adds failing test for importing a regular .sass partial

kennethormandy:
Here’s the issue [I mentioned](https://twitter.com/kennethormandy/status/684961843744387072) on Twitter. After testing it a bit more, it seems like Eyeglass is just interfering with or not processing with files with `.sass` indented syntax partials:

__Expected__

I can import a `.sass` and/or `.scss` partials, from `.sass` or `.scss` files, while using Eyeglass with Node-sass.

__Actual__

I can import `.scss` partials from `.sass` or `.scss` files, while using Eyeglass with Node-sass. If I import a `.sass` partial from either a `.sass` file or a `.scss` file, there is not CSS outputted from Node-sass.

This pull request adds one passing test for `.scss`, and one failing test for `.sass` demonstrating this. Hopefully it is just some oversight on my part regarding `.sass` syntax, but I can’t seem to get around this.

### DIFF
--- a/test/fixtures/sass_indented_import/sass/_part.sass
+++ b/test/fixtures/sass_indented_import/sass/_part.sass
@@ -1,0 +1,2 @@
+body
+  background: #FFF
--- a/test/fixtures/sass_indented_import/sass/main.sass
+++ b/test/fixtures/sass_indented_import/sass/main.sass
@@ -1,0 +1,4 @@
+@import "part"
+
+h1
+  font-family: Georgia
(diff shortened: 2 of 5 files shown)


eoneill:
Probably related to https://github.com/sass/node-sass/issues/1192

chriseppstein:
Merged, but I disabled the failing test for now in commit 0b7f9f8bb37ef21a9a4fad8bd7f482080ef85190

PR #118, opened 2016-01-08
allow eyeglass to be imported from anywhere

eoneill:
This issue was uncovered in https://github.com/sass-eyeglass/eyeglass-spriting/pull/19 where `eyeglass-spriting` tries to `@import "eyeglass/assets"` but doesn't declare `eyeglass` as a direct dependency.


chriseppstein:
:+1:

PR #120, opened 2016-01-12
only load main if isEyeglassModule

### LINKED ISSUE #119: Code broke after change from 0.8.0 -> 0.8.1
Hello, I use eyegalss with gulp, using what (I think is) pretty simple setup.
```
gulp.task('sass', function() {
    return gulp.src('./src/sass/main.scss')
        .pipe(sass(eyeglass(nodeSassOpts)).on('error', sass.logError))
        .pipe(gulp.dest('./dist/css'))
        .pipe(livereload());
});
```
And my code worked good on version 0.8, but with the latest change it doesn't work.
It's trying to load the file 'index.js' from my root directory, which i don't include anywhere and since there is no such file it fails.
I don't know what the problem might be, yet i see in newest commits that there were some changes regarding file reading.

I double checked that sass alone (without eyeglass) compiles without problems, and that the problem occurs only on 0.8.1 and not on 0.8.0

eoneill:
In eyeglass 0.8.1, we incorrectly `require()` the root projects `main` even if it's not an `eyeglass-module`.

This ensures that it's an `eyeglass-module` before we `require(moduleMain)`.

This should resolve #119.


chriseppstein:
:+1:

PR #123, opened 2016-01-23
fix module importer pulling in incorrect module index when file does not exist

eoneill:
This is a pretty nasty bug where the importer will actually import a file that has nothing to do with the import uri.

For example, assume in `module/foo.scss`...
```scss
@import "does-not-exist";
```

Will actually result in `module/_index.scss` or similar variant getting imported. The current logic in `expandFileName` adds numerous files that shouldn't be included, as well as incorrect ordering.

The previous order was:
```js
[
  'foo/bar',
  'foo/bar.scss',
  'foo/bar/index.scss',
  'foo/bar.sass',
  'foo/bar/index.sass',
  'foo/bar.css',
  'foo/bar/index.css',
  'foo/_bar.scss',
  'foo/bar/_index.scss',
  'foo/_bar.sass',
  'foo/bar/_index.sass',
  'foo/_bar.css',
  'foo/bar/_index.css'
]
```

The ordering with this update is now:
```js
[
  'foo/bar',
  'foo/bar.scss',
  'foo/bar.sass',
  'foo/bar.css',
  'foo/_bar.scss',
  'foo/_bar.sass',
  'foo/_bar.css',
  'foo/bar/index.scss',
  'foo/bar/index.sass',
  'foo/bar/index.css',
  'foo/bar/_index.scss',
  'foo/bar/_index.sass',
  'foo/bar/_index.css'
]
```

This also departs from the existing implementation that ignores the possible extensions if the uri already has an extension on it.

e.g.
```
@import "foo.css";
```

would only map to `foo.css` and `_foo.css`. The updated logic will now map to `foo.css.scss`, `foo.css/index.scss`, etc. See new `test_NameExpander` test case.


chriseppstein:
If an extension is specified, only that extension should be considered in the lookup logic.

test/test_NameExpander.js
```diff
@@ -0,0 +1,68 @@
...
+      "test-uri.css",
+      "_test-uri.scss",
+      "_test-uri.sass",
+      "_test-uri.css",
+      "test-uri/index.scss",
+      "test-uri/index.sass",
+      "test-uri/index.css",
+      "test-uri/_index.scss",
+      "test-uri/_index.sass",
+      "test-uri/_index.css"
```
chriseppstein: The [Ruby Sass importer logic](https://github.com/sass/sass/blob/stable/lib/sass/importers/filesystem.rb#L118-L122) specifies that if an extension is specified, only that extension is considered. So this change is not desired.

It also specifies that if the extension is `.css` then the sass implementation will create a CSS-based import to a relative URI, so our importer should actually never be invoked if the uri being `@import`ed is `foo.css`. If our importer is being called, this is a libsass bug. In Sass 4.0 we're changing to a different directive (`@use`) which will allow for css being invoked like this, so it's ok for our name expander to handle this case, just that we should know it won't actually happen in an integration test.

test/test_NameExpander.js
```diff
@@ -0,0 +1,68 @@
...
+      "test-uri.css.css",
+      "_test-uri.css.scss",
+      "_test-uri.css.sass",
+      "_test-uri.css.css",
+      "test-uri.css/index.scss",
+      "test-uri.css/index.sass",
+      "test-uri.css/index.css",
+      "test-uri.css/_index.scss",
+      "test-uri.css/_index.sass",
+      "test-uri.css/_index.css"
```
chriseppstein: The [Ruby Sass importer logic](https://github.com/sass/sass/blob/stable/lib/sass/importers/filesystem.rb#L118-L122) specifies that if an extension is specified, only that extension is considered. So this change is not desired.

It also specifies that if the extension is `.css` then the sass implementation will create a CSS-based import to a relative URI, so our importer should actually never be invoked if the uri being `@import`ed is `foo.css`. If our importer is being called, this is a libsass bug. In Sass 4.0 we're changing to a different directive (`@use`) which will allow for css being invoked like this, so it's ok for our name expander to handle this case, just that we should know it won't actually happen in an integration test.

chriseppstein: If the uri is `"test-uri.scss"` then only `"test-uri.scss"` and `"_test-uri.scss"` should be considered.

chriseppstein:
Looking now, If the import uri is `foo/bar` then a file with no extension should not be considered. Ruby sass wouldn't load that file.

The new import resolution order is fine. preferable I think.

test/test_NameExpander.js
eoneill: Hrmm. I was following the way libSass implemented this. Without eyeglass, if you `@import "foo.css"`, `foo.css.scss` is a valid file to import.

I'll update the logic to follow the Ruby Sass implementation.

eoneill:
Thanks for the feedback. I'll make the necessary updates today.

lib/util/eyeglass_module.js
```diff
@@ -38,6 +39,11 @@ function EyeglassModule(mod, discoverModules, isRoot) {
         main: moduleMain && require(moduleMain),
         mainPath: moduleMain
       });
+
+      if (rInvalidName.test(mod.name)) {
+        console.log("invalid name");
```
chriseppstein: left over from debugging?

eoneill: That was already removed. Stop digging up mistakes from my past ;)

Issue #126, opened 2016-02-02
Eyeglass modules not discovered if used as peerDependency


archnode:
If I use an eyeglass module solely as a peer dependency in the projects package.json, the module won't get included properly. Adding the same package as a regular dependency fixes the problem.

It would be great if peerDependencies could have the same behavior as other dependencies.

Thanks for your work on this!

eoneill:
@chriseppstein what are your thoughts? I think this makes sense.

Open question: do we treat `peerDependencies` like normal `dependencies`, or do we treat them like `devDependencies`? That is, do we only look at `peerDependencies` for the top-level (root)?

eoneill:
I think the right call is to always include `peerDependencies` even in sub-modules. Discuss in #127

PR #127, opened 2016-02-02
add peerDependencies to module discovery (#126)

eoneill:
Per #126, adds support for discovering modules declared in `peerDependencies`.


chriseppstein:
This needs a test case for `peerDependencies` declared in an eyeglass module's `package.json`. I'm not sure whether that has the peer dep installed into the node_modules of the module or the project itself.

eoneill:
Added a test case with the nested peerDependency.

This "magically" works because we add it to the dependency hierarchy. When we invoke `eyeglass.modules.access(...)` [in the importer](https://github.com/sass-eyeglass/eyeglass/blob/master/lib/module_importer.js#L79), it correctly resolves the deduped module, even though the module lives in a parent `node_module` directory.

chriseppstein:
:+1:

Issue #128, opened 2016-02-05
Eyeglass modules not discovered on Windows


JohnRiv:
Testing on Windows 10, I noticed that my eyeglass modules that import automatically on my Mac are not imported on Windows. I've narrowed it down to [this path comparison](https://github.com/sass-eyeglass/eyeglass/blob/e8f4161924e24e8284ff2f0733ade9d9cbb15deb/lib/util/eyeglass_modules.js#L251):

```js
// if the module path matches the search path, push it onto our results
if (mod.path === dir) {
  branches.push(mod);
}
```

On a Mac, these are identical, e.g.:
mod.path: `/Users/johnriv/projects/eyeglass-test`
dir: `/Users/johnriv/projects/eyeglass-test`

However on Windows, they're different:
mod.path: `C:\Users\IEUser\Documents\eyeglass-test`
dir: `C:/Users/IEUser/Documents/eyeglass-test`

eoneill:
Thanks for reporting this. I suspected this last night as I was working on some cleanup. We're not very consistent with the way we handle uris and file paths and I started writing some helpers to normalize this.

I'm cleaning stuff up for assets right now, and should be able to work this in.

JohnRiv:
Sounds great, thanks. Let me know when the code is pushed up and I can test it out.

eoneill:
I've got most things working now on Windows (tested on Win10) with the latest changes in my branch: https://github.com/eoneill/eyeglass/tree/assets-refactor

I still have a bit more cleanup to do, but things should work on Windows soon.

Will update when I submit a pull request.

JohnRiv:
Thanks for the update @eoneill. I tried out your progress so far and it works on my basic Eyeglass test project. With eyeglass-spriting though, I'm running into this error on Windows 10:

```
C:\Users\IEUser\Documents\eyeglass-test>grunt css
Running "sass:dist" (sass) task
>> error in C function eyeglass-asset-uri: Asset not found: spritemaps\icon-sprite-map.png
>>
>> Backtrace:
>>      node_modules/eyeglass/sass/assets.scss:48, in function `eyeglass-asset-uri`
>>      node_modules/eyeglass/sass/assets.scss:48, in function `asset-uri`
>>      node_modules/eyeglass/sass/assets.scss:53, in function `asset-url`
>>      node_modules/eyeglass-spriting/sass/index.scss:17, in function `sprite-url`
[...]

Issue #79, opened 2015-11-13
write quick start guide


xiwcx:
Based off of @chriseppstein's presentation at the Sass unconference, write a guide for people who just want to get eyeglass up and running.

arkhi:
That would be great indeed. :)

I’m struggling to make eyeglass to integrate with grunt and grunt-sass, based on the different examples found on different websites. It seems the documentation is not in sync, so I can’t tell if something is deprecated or if there is just different ways to do the same thing.

[On eyeglass.rocks]():
```js
sass: {
    options: require("eyeglass").Eyeglass({
        sourceMap: true
    }).sassOptions(),
```

[On npmjs.com](https://www.npmjs.com/package/eyeglass#example-integration-with-grunt-and-grunt-sass):
```js
var eyeglass = require("eyeglass");
...
sass: {
    options: eyeglass.decorate({
        sourceMap: true
    }),
```
[On github.com](https://github.com/sass-eyeglass/eyeglass#example-integration-with-grunt-and-grunt-sass):
```js
var eyeglass = require("eyeglass");
...
sass: {
    options: eyeglass({sourceMap: true}),
```

eoneill:
@arkhi, the API is changing and this may cause some confusion with existing documentation.

If you're using 0.7.x (current release is 0.7.1) or earlier, you're going to use either:

```js
var eyeglass = require("eyeglass");
var options = eyeglass.decorate(...);

// OR

var Eyeglass = require("eyeglass").Eyeglass;
var eyeglass = new Eyeglass(...);
var options = eyeglass.sassOptions();
```
When 0.8.0 is released, the above will be deprecated (they'll still work, but with warnings).
The new API simplifies it a bit to...
```js
var options = require("eyeglass")(...);

// OR

var Eyeglass = require("eyeglass");
var eyeglass = new Eyeglass(...);
var options = eyeglass.options;
```

So in short, npmjs.org and eyeglass.rocks are both accurate for the current release. The docs here on github (which reflect HEAD) are accurate for the next release.

arkhi:
@eoneill: I see… Thanks a lot for the explanation.

I still have some glitches with variables imported from other partials, after including eyeglass 0.7.1 in the loop. I don’t know if it’s relevant but I’ll fill a bug for it.

georeith:
[...]

PR #129, opened 2016-02-09
refactor assets, assets importer, and fix Windows paths

eoneill:
This is a substantial refactor.

This pull request contains the following fixes / changes:
- `@import "assets"` within a module does not currently work
  - This was addressed in part by #123
- The root directory is always treated as an "app", even if the root directory is actually a module (`isEyeglassModule === true`).
  - This leads to inconsistencies in assets in the current project vs when the project is consumed as a dependency.
  - See https://github.com/sass-eyeglass/eyeglass/issues/79#issuecomment-173171567
- When using `httpRoot` with `relativeTo`, incorrect asset paths are generated.
- Windows paths separators and delimiters are used inconsistently throughout eyeglass. This takes the first step towards addressing this by introducing a new set of uri utility methods.
  - See #128
- This significantly reorganizes the project structure for better maintainability.

```
lib
├── assets
│   ├── Assets.js
│   ├── AssetsCollection.js
│   └── AssetsSource.js
├── eyeglass-exports.js
├── functions
│   ├── asset-uri.js
│   ├── index.js
│   └── version.js
├── importers
│   ├── AssetImporter.js
│   ├── ImportUtilities.js
│   └── ModuleImporter.js
├── index.js
├── modules
│   ├── EyeglassModule.js
│   ├── EyeglassModules.js
│   └── ModuleFunctions.js
└── util
    ├── NameExpander.js
    ├── Options.js
    ├── debug.js
    ├── deprecator.js
    ├── files.js
    ├── package.js
    ├── resolve.js
    ├── semverChecker.js
    ├── strings.js
    ├── sync.js
    └── uri.js
```

To see individual commit history, please see: https://github.com/eoneill/eyeglass/commits/assets-refactor


lib/util/NameExpander.js
```diff
@@ -24,7 +24,7 @@ function NameExpander(uri) {
   * @param   {String} location - the location path the expand the URI against
   */
 NameExpander.prototype.addLocation = function(location) {
-  if (location && location !== "stdin" && fileUtils.existsSync(location)) {
+  if (location && location !== "stdin") {
```
eoneill: TODO - I need to add a test for this

ryuran:
Some news about this pull request ?

codegecko:
An update would be appreciated, trying to use bootstrap-sass with Eyeglass on Windows still isn't working and I really need it to (W10 x64 dev here), and trying to check out the git repo with the specific amends fails cause eyeglass's version checking doesn't recognise npm git-based dependencies.

Snugug:
I've run the fork at it's working as expected in Windows. As per @eoneill's instructions, this is what I did:

* Changed my Eyeglass source in `package.json` to this fork/branch (`eoneill/eyeglass#assets-refactor-cleaner`)
* Added `process.env.EYEGLASS_NORMALIZE_PATHS = true;` to my Node file where Eyeglass is run from

After doing those two things, paths on Windows work again.

We've got this working for a bunch of Windows machines; it'd be great if this could make it in so we don't need to point to the fork.

chriseppstein:
Closing. This landed in a different PR.

PR #135, opened 2016-07-01
README clearification

xi:
This was keeping me much to long from using eyeglass! After reading the README, I thought that eyeglass would automatically hook into node-sass. I thought that my modules where not discovered for some reason. It took me ages to understand that the issue was that node-sass is not "eyeglass-aware".

In short: I think this change makes it much easier for people to start using eyeglass. Kind of related: https://github.com/sass/node-sass/pull/1617

**Note**: The link only works on github. Should I change it to an absolute URL?


chriseppstein:
Thanks!

PR #137, opened 2016-09-01
[updated] refactor assets, assets importer, and fix Windows paths

eoneill:
This is a cleaned up branch that takes precedence over #129. See #129 for discussion.

Additional changes here include:
- URI path normalization is enabled by default (with an opt-out flag/option, see README).
- Literal backslash (`\`) in `asset-uri` and `asset-url` are not normalized. This means that asset references in CSS should always use Unix style (`/`) path separators.


chriseppstein:
Merged and released as 0.9.0

PR #145, opened 2016-09-29
fixes for libsass 3.4 support

eoneill:
The main change here is to not `quote()` the output directly in `eyeglass-normalize-uri`. This caused issues when the resulting value was set as a key on `$eg-registered-assets`. See https://github.com/sass/libsass/issues/2197

Minor updates to `assets.scss` and tests to account for this change in the output.

Verified tests passing with...

```
node-sass	3.10.0	(Wrapper)	[JavaScript]
libsass  	3.3.6	(Sass Compiler)	[C/C++]
```
and
```
node-sass	3.11.0	(Wrapper)	[JavaScript]
libsass  	3.4.0	(Sass Compiler)	[C/C++]
```

FWIW, this should have been resolvable by `unquote`ing the output when setting the key, but this still resulted in a quoted key, so this might still be an issue with `unquote` in libsass.

```scss
$eg-registered-assets: map-merge($eg-registered-assets, (
  $module: map-merge($mod-assets, (
    unquote($url): ...
```

/cc @xzyfer @chriseppstein

### DIFF
--- a/lib/functions/normalize-uri.js
+++ b/lib/functions/normalize-uri.js
@@ -9,8 +9,6 @@ module.exports = function(eyeglass, sass) {
       var uri = $uri.getValue();
       // normalize the uri for the given type
       uri = URI[type](uri);
-      // then normalize it to an escaped sass string
-      uri = URI.sass(uri);
       done(sass.types.String(uri));
     },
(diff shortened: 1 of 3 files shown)


xzyfer:
\o/

PR #162, opened 2017-02-14
Add gulp example to Readme

nfriedly:
This was pretty simple but still took me a few minutes of poking around to figure out, partially because there are still references to the older `new Eyeglass().sassOptions()` API that I was trying to use initially.

### DIFF
--- a/README.md
+++ b/README.md
@@ -460,7 +460,7 @@ var eyeglass = require("eyeglass");
  
 gulp.task('sass', function () {
   return gulp.src('./sass/**/*.scss')
-    .pipe(sass(eyeglass({sourceMap: true}).on('error', sass.logError))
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -451,6 +451,20 @@ sass: {
...
+### Example: integration with gulp and gulp-sass
+
+```js
+var gulp = require('gulp');
+var sass = require('gulp-sass');
+var eyeglass = require("eyeglass");
+ 
+gulp.task('sass', function () {
+  return gulp.src('./sass/**/*.scss')
+    .pipe(sass(eyeglass({sourceMap: true}).on('error', sass.logError))
```
jackw: Looks like there's a missing closing bracket before `.on('error', sass.logError))`.

nfriedly: good catch, fixed now.

PR #172, opened 2017-06-21
better support for builds that use symlink

eoneill:
This issue was discovered while testing with broccoli 1.0. By default, broccoli will now use a global `tmp` path rather than local. This means that files get symlink'd in e.g. `/var/folders/...`

When eyeglass attempts to do it's module `access` check, it walks up the directory tree until it finds a `package.json`. It then uses that `package.json` to determine what dependencies can be imported.

When these files are symlink'd outside of the project workspace, eyeglass fails to find a `package.json` and thereby determines the origin scss file does not have access to the requested modules.

The fix here is to follow the symlink and check to see if the "real" file has access.

Note that we have to test _both_ the real and symlink paths. We have to test the original symlink path because it's valid to have symlink'd files _within_ a project and they should check access permission relative to the project before following the symlink.

Did not add tests because this is hard to test. Reduced test coverage thresholds a bit as such.

### DIFF
--- a/lib/modules/EyeglassModules.js
+++ b/lib/modules/EyeglassModules.js
@@ -8,6 +8,7 @@ var path = require("path");
 var merge = require("lodash.merge");
 var semver = require("semver");
 var archy = require("archy");
+var fs = require("fs");
 var URI = require("../util/URI");
 
 /**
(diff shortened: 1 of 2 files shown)


lib/modules/EyeglassModules.js
```diff
@@ -253,39 +254,57 @@ function canAccessModule(name, origin) {
...
 
-  // cache it for future lookups
-  cache[name][pkg] = canAccess;
+  // check if we can access from the origin...
+  var canAccess = canAccessFrom(origin);
+
+  // if not...
+  if (!canAccess) {
+    // check for a symlink...
+    var realOrigin = fs.realpathSync(origin);
```
stefanpenner: is there a reason not to always `realpathSync` ? node's resolution algorithm takes that approach.

stefanpenner:
I don't believe this has been released yet, any reason to hold off?

lib/modules/EyeglassModules.js
chriseppstein: This could make a file that is linked to appear like it belongs to a particular module no longer appear as part of the module. It's possible that our validations are being done at the wrong time against the wrong values and that fixing the way those work would be the best solution -- but I think using realpath all the time implies more testing and code changes.

stefanpenner: > but I think using realpath all the time implies more testing and code changes.

Makes sense.  I suspect this may be something todo, as it (from how I understand eyeglass to work) work more like the node `require`, and may lead to less surprises when symlinking. it's also possible I don't quite understand this part of eyeglass, and this doesn't make sense.

PR #178, opened 2018-03-13
Cache scss generated to register assets

mikrostew:
## Problem

Every time `@import "module/assets"` is rendered, the code to register the assets is generated from scratch. In a large codebase where the same modules are used 100s of times, this wastes a lot of time doing the same thing over and over.

## Solution

Cache the generated code the first time, and return that for any subsequent `@import`s that are using the same module and options.

### DIFF
--- a/lib/assets/Assets.js
+++ b/lib/assets/Assets.js
@@ -200,6 +200,13 @@ Assets.prototype.installer = function(installer) {
   };
 };
 
+/**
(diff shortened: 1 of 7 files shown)


mikrostew:
As implemented, this is a breaking change, which is no bueno. Dropped in favor of https://github.com/sass-eyeglass/eyeglass/pull/179

PR #179, opened 2018-03-15
Generate cache keys for sources and collections

mikrostew:
Adds a few things to allow build tools to cache the Scss generated in asAssetImport(), since otherwise this is done over and over:

- keeps a list of references to the AssetsCollections produced for modules
- adds functions to produce cache keys for AssetsSource and AssetsCollection

### DIFF
--- a/lib/assets/Assets.js
+++ b/lib/assets/Assets.js
@@ -14,6 +14,8 @@ function Assets(eyeglass, sass) {
   this.eyeglass = eyeglass;
   // create a master collection
   this.collection = new AssetsCollection();
+  // and keep a list of module collections
+  this.moduleCollections = [];
 
   // Expose these temporarily for back-compat reasons
   function deprecate(method) {
@@ -57,6 +59,7 @@ Assets.prototype.addSource = function(src, opts) {
   */
 Assets.prototype.export = function(src, opts) {
   var assets = new AssetsCollection();
+  this.moduleCollections.push(assets);
   return assets.addSource(src, opts);
 };
 
--- a/lib/assets/AssetsCollection.js
+++ b/lib/assets/AssetsCollection.js
@@ -57,9 +57,9 @@ AssetsCollection.prototype.asAssetImport = function (name) {
   * @returns {String} the cache key
   */
 AssetsCollection.prototype.cacheKey = function(name) {
-  return this.sources.reduce(function(cacheStr, source) {
-    return cacheStr + ":" + source.cacheKey(name);
-  }, "sources");
+  return this.sources.map(function(source) {
+    return source.cacheKey(name);
+  }).sort().join(":");
 };
 
 module.exports = AssetsCollection;
--- a/lib/assets/AssetsCollection.js
+++ b/lib/assets/AssetsCollection.js
@@ -52,4 +52,14 @@ AssetsCollection.prototype.asAssetImport = function (name) {
   }, '@import "eyeglass/assets";\n');
 };
 
+/**
+  * Build a string suitable for caching an instance of this
+  * @returns {String} the cache key
+  */
+AssetsCollection.prototype.cacheKey = function(name) {
+  return this.sources.reduce(function(cacheStr, source) {
+    return cacheStr + ":" + source.cacheKey(name);
+  }, "sources");
+};
+
 module.exports = AssetsCollection;
--- a/lib/assets/AssetsSource.js
+++ b/lib/assets/AssetsSource.js
@@ -6,6 +6,7 @@ var path = require("path");
 var merge = require("lodash.merge");
 var URI = require("../util/URI");
 var fileUtils = require("../util/files");
+var stringify = require("json-stable-stringify");
 
 /* class AssetsSource
  *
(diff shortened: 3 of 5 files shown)


lib/assets/AssetsSource.js
```diff
@@ -82,4 +82,32 @@ AssetsSource.prototype.toString = function() {
   return this.srcPath + "/" + this.pattern;
 };
 
+/**
+  * Build a string suitable for caching an instance of this
+  * @returns {String} the cache key
+  */
+AssetsSource.prototype.cacheKey = function(namespace) {
+  function skipSomeKeys(key, value) {
```
stefanpenner: I don't believe this closes over anything useful, so lets hoist this out to the root of the module, this will prevent the necessary redefinition of `skipSomeKeys` each time `cacheKey` function is invoked

test/test_assets.js
```diff
@@ -983,6 +984,122 @@ describe("assets", function () {
...
+          });
+          var source2 = new AssetsSource(rootDir, {
+            httpPrefix: "foo",
+          });
+          var source3 = new AssetsSource(rootDir, {
+            httpPrefix: "bar",
+          });
+          assert.equal(source1.cacheKey(), source2.cacheKey());
+          assert.notEqual(source1.cacheKey(), source3.cacheKey());
+          done();
```
stefanpenner: In mocha, we should likely not use `done()`, two reasons:

1. sync tests, such as this don't require it. It can safely be dropped.
2. async tests, should preferrably rely on a promise being returned in the `it` `function` body, as it does error handling correctly (timed out tests do to test failures suck)

test/test_assets.js
```diff
@@ -983,6 +984,122 @@ describe("assets", function () {
...
+          });
+          var source2 = new AssetsSource(rootDir, {
+            name: "foo",
+          });
+          var source3 = new AssetsSource(rootDir, {
+            name: "bar",
+          });
+          assert.equal(source1.cacheKey(), source2.cacheKey());
+          assert.notEqual(source1.cacheKey(), source3.cacheKey());
+          done();
```
stefanpenner: ^

test/test_assets.js
```diff
@@ -983,6 +984,122 @@ describe("assets", function () {
...
+        });
+
+        it("Assets.cacheKey includes namespace", function(done) {
+          var source1 = new AssetsSource(rootDir, {});
+          var source2 = new AssetsSource(rootDir, {});
+          assert.equal(source1.cacheKey("foo"), source2.cacheKey("foo"));
+          assert.notEqual(source1.cacheKey("foo"), source2.cacheKey("bar"));
+          done();
```
stefanpenner: ^

test/test_assets.js
```diff
@@ -983,6 +984,122 @@ describe("assets", function () {
...
+        });
+
+        it("Assets.cacheKey includes srcPath", function(done) {
+          var source1 = new AssetsSource(rootDir, {});
+          var source2 = new AssetsSource(rootDir, {});
+          var source3 = new AssetsSource(rootDir2, {});
+          assert.equal(source1.cacheKey(), source2.cacheKey());
+          assert.notEqual(source1.cacheKey(), source3.cacheKey());
+          done();
```
stefanpenner: ^

test/test_assets.js
```diff
@@ -983,6 +984,122 @@ describe("assets", function () {
...
+          });
+          var source2 = new AssetsSource(rootDir, {
+            pattern: "images/**/*",
+          });
+          var source3 = new AssetsSource(rootDir, {
+            pattern: "images/**/*.jpg",
+          });
+          assert.equal(source1.cacheKey(), source2.cacheKey());
+          assert.notEqual(source1.cacheKey(), source3.cacheKey());
+          done();
```
stefanpenner: ^

test/test_assets.js
```diff
@@ -983,6 +984,122 @@ describe("assets", function () {
...
+          var source1 = new AssetsSource(rootDir, {});
+          var source2 = new AssetsSource(rootDir, {});
+          var source3 = new AssetsSource(rootDir, {
+            globOpts: {
+              dot: true
+            }
+          });
+          assert.equal(source1.cacheKey(), source2.cacheKey());
+          assert.notEqual(source1.cacheKey(), source3.cacheKey());
+          done();
```
stefanpenner: ^

test/test_assets.js
```diff
@@ -983,6 +984,122 @@ describe("assets", function () {
...
+            }
+          });
+          var source3 = new AssetsSource(rootDir, {
+            globOpts: {
+              nonegate: true
+            }
+          });
+          assert.equal(source1.cacheKey(), source2.cacheKey());
+          assert.equal(source1.cacheKey(), source3.cacheKey());
+          done();
```
stefanpenner: ^

test/test_assets.js
```diff
@@ -983,6 +984,122 @@ describe("assets", function () {
...
+          collection1.addSource(rootDir);
+          collection2.addSource(rootDir);
+          collection3.addSource(rootDir2);
+          collection4.addSource(rootDir);
+          collection4.addSource(rootDir2);
+
+          assert.equal(collection1.cacheKey(), collection2.cacheKey());
+          assert.notEqual(collection1.cacheKey(), collection3.cacheKey());
+          assert.notEqual(collection1.cacheKey(), collection4.cacheKey());
+          done();
```
stefanpenner: ^

lib/assets/AssetsSource.js
```diff
@@ -82,4 +82,32 @@ AssetsSource.prototype.toString = function() {
...
+    }
+    return value;
+  }
+
+  return "[" +
+    "httpPrefix=" + (this.httpPrefix ? this.httpPrefix : "") +
+    ";name=" + (this.name ? this.name : (namespace ? namespace : "")) +
+    ";srcPath=" + this.srcPath +
+    ";pattern=" + this.pattern +
+    ";opts=" + JSON.stringify(this.globOpts, skipSomeKeys) +
```
stefanpenner: `JSON.stringify` isn't stable, so it will sometimes produce different ordering of keys in the output. I would recommend using `json-stable-stringify` instead.

lib/assets/Assets.js
```diff
@@ -57,6 +59,7 @@ Assets.prototype.addSource = function(src, opts) {
   */
 Assets.prototype.export = function(src, opts) {
   var assets = new AssetsCollection();
+  this.moduleCollections.push(assets);
```
stefanpenner: Where is this used? If it is new public API, we should be sure to add appropriate tests so that it doesn't vanish.

lib/assets/AssetsCollection.js
```diff
@@ -52,4 +52,14 @@ AssetsCollection.prototype.asAssetImport = function (name) {
   }, '@import "eyeglass/assets";\n');
 };
 
+/**
+  * Build a string suitable for caching an instance of this
+  * @returns {String} the cache key
+  */
+AssetsCollection.prototype.cacheKey = function(name) {
+  return this.sources.reduce(function(cacheStr, source) {
```
stefanpenner: should we `sort` `sources` to ensure they remain stable for cacheKey generation?

test/test_assets.js
mikrostew: ok, I'll remove the use of done() in all these places

lib/assets/AssetsSource.js
mikrostew: oh ok, I didn't know that - I'll change that

lib/assets/Assets.js
mikrostew: yes, this is new API to have access to the AssetsCollections in modules (for caching them)

lib/assets/AssetsCollection.js
mikrostew: yup, good catch

lib/assets/AssetsSource.js
stefanpenner: ya, i found out the hard way...

stefanpenner:
released as v1.4.0 🎉

PR #181, opened 2018-04-04
normalize-uri isn’t required in posix, and is quite slow

stefanpenner:


test/test_assets.js
```diff
@@ -295,7 +295,7 @@ describe("assets", function () {
           "$uri: 'assets/foo/bar.png');" +
           ".test { foo: inspect($eg-registered-assets); }";
```
chriseppstein: change this line to `.test { foo: inspect(map-get(map-get($eg-registered-assets, module-a), 'foo/bar.png')); }` and adjust the output accordingly.

stefanpenner:
```
¦.test {
  ¦ ¦ -  foo: (module-a: ("foo/bar.png": (filepath: "images/foo/bar.png", uri: "assets/foo/bar.png"))); }
  ¦ ¦ +  foo: (filepath: "images/foo/bar.png", uri: "assets/foo/bar.png"); }
```

chriseppstein: the new output looks good, we can set the expected value to that and the tests should pass now, yes?

stefanpenner:
Without: `EyeglassCompiler (33) | 137006ms (4151 ms)`
With: `EyeglassCompiler (33) | 95998ms (2909 ms)`

chriseppstein:
@eoneill  can you review?

stefanpenner:
```
eoneill [8:12 AM]
On DTO and limited internet ATM. Looks fine, but can you fix the doc block?

spenner [8:15 AM]
which doc block (sure)

eoneill [8:17 AM]
Looks like the new sass function (eyeglass-normalize-uri) in scss got added under the doc block for the old function
Looks like the new sass function (eyeglass-normalize-uri) in scss got added under the doc block for the old function
Sorry on subway, internet is cutting in and out

spenner [8:18 AM]
ohohoh
i see
yup will fix

eoneill [8:19 AM]
:+1:
```

PR #186, opened 2018-09-12
POC cache modules, the same way node’s require caches modules.

stefanpenner:
This yields a pretty non-trivial improvement. But obviously needs to be implemented more carefully. Today (in this spike) the `this` in `resolveModule` is not properly considered, and may lead to issue…

Opening this to further conversation on the topic.

- [ ] ensure this doesn't actually screw anything up
- [ ] implement properly
- [ ] tests
- [ ] cache purge, maybe something similar to `require.entries`


chriseppstein:
This is a good discovery but I'd prefer to land this perf win with a different design. I'll file a new issue to track this.

Issue #187, opened 2018-09-12
Cache module discovery


chriseppstein:
There's a bunch of time being wasted by doing repeated module discovery. Because we use require to access package.json files, we can cache module dependency information globally for the duration of the process (and should for consistency). Alternatively, we should not use require and create a different cache context.

https://github.com/sass-eyeglass/eyeglass/pull/186

eoneill:
resolved with #189

PR #189, opened 2018-10-02
add module discovery caching per #187

eoneill:
This adds module discovery caching to `EyeglassModules`.

The cache is fairly simple and is stored per instance of `EyeglassModules` (which translates to per `Eyeglass` instance).

I tested against a fairly large code base to see the difference between per-instance vs global caching and the savings were statistically insignificant. That said, if you are creating hundreds/thousands of eyeglass instances, it might be beneficial.

Here are some metrics of time spent in `discoverModules` (averaged across 3 runs each):
- **no caching**: 114.349s
- **per-instance caching**: 34.372s
- **global caching**: 37.052s

Other minor changes:
- updated the `canAccess` cache to share the same `SimpleCache`
- updated `debug.*` to only create functions if the debugger is enabled. this allows us to short-circuit some semi expensive debug message generation if we aren't going to use it.

Related: #187 #186

### DIFF
--- a/lib/modules/EyeglassModules.js
+++ b/lib/modules/EyeglassModules.js
@@ -12,14 +12,17 @@ var archy = require("archy");
 var fs = require("fs");
 var URI = require("../util/URI");
 
+var globalModuleCache = new SimpleCache();
+
 /**
   * Discovers all of the modules for a given directory
   *
   * @constructor
   * @param   {String} dir - the directory to discover modules in
   * @param   {Array} modules - the explicit modules to include
+  * @param   {Boolean} useGlobalModuleCache - whether or not to use the global module cache
   */
-function EyeglassModules(dir, modules) {
+function EyeglassModules(dir, modules, useGlobalModuleCache) {
   this.issues = {
     dependencies: {
       versions: [],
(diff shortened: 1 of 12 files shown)


lib/modules/EyeglassModules.js
```diff
@@ -257,39 +260,34 @@ function canAccessModule(name, origin) {
...
-      dependencies: this.tree
-    }, pkg);
-
-    var canAccess = branches.some(function(branch) {
-      // if the reference is to itself (branch.name)
-      // OR it's an immediate dependency (branch.dependencies[name])
-      if (branch.name === name || branch.dependencies && branch.dependencies[name]) {
-        return true;
-      }
+    var cacheKey = JSON.stringify({
```
stefanpenner: stringify isn't stable by default, the cache key may not result in collisions due to this.

would `${origin}|${pkg.name}|${pkg.version}` be sufficient, or better yet `realpath` of that location?

If not, maybe [the following lib](https://github.com/substack/json-stable-stringify) may be good.

stefanpenner:
> global caching: 37.052s

Do you happen to have a breakdown of the remaining cost centers?

eoneill:
Previous metrics were not accurate due to double counting.

Here are updated metrics (measuring the top-level `resolveModule` call in `EyeglassModules`):
- **no caching**: 78.639787s
- **per-instance caching**: 35.079668s
- **global caching**: 0.176994s

Raw metrics:
https://docs.google.com/spreadsheets/d/1ac0Z_hSWjf16UWlFzrTX17CKm7uq9hcaegegzSJjUmU/edit?usp=sharing

Issue #183, opened 2018-05-31
eyeglass.rocks doesn't work


toddself:
you get a github 404 page.

Jakobo:
We seem to be missing the gh-pages branch. It may be as simple as re-running the [gulp deployer](https://github.com/sass-eyeglass/eyeglass/blob/master/build/site.js) for the site. I'll give it a try this weekend and see if I can get it working. Thanks for flagging it, and apologies for the long delay in picking up this bug.

Jakobo:
Just an update. I'll put a separate PR (#190) up for the changes needed to get the gulp deployer working. It generates the `gh-pages` branch correctly. However, we need to update the `A` record.

> The custom domain for your GitHub Pages site is pointed at an outdated IP address. You must update your site's DNS records if you'd like it to be available via your custom domain. For more information, see https://help.github.com/articles/setting-up-a-custom-domain-with-github-pages/.

@chriseppstein, do you own the DNS record for eyeglass.rocks? We should be able to get that fixed and bring the docs pages back.

stefanpenner:
@Jakobo I believe the DNS seems correct:

```
dig eyeglass.rocks

; <<>> DiG 9.10.6 <<>> eyeglass.rocks
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 19340
;; flags: qr rd ra; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4000
;; QUESTION SECTION:
;eyeglass.rocks.			IN	A

;; ANSWER SECTION:
eyeglass.rocks.		253	IN	A	192.30.252.153
eyeglass.rocks.		253	IN	A	192.30.252.154

;; Query time: 0 msec
;; SERVER: 2620:119:5001:4f01::106#53(2620:119:5001:4f01::106)
;; WHEN: Fri Nov 02 17:09:13 PDT 2018
;; MSG SIZE  rcvd: 75
```

And 192.30.252.153 appears to be the correct GH pages IP address right? Or did I misunderstand.

Jakobo:
Looks like eyeglass.rocks is up. We can always create a separate issue for the DNS. Thanks @stefanpenner.

PR #190, opened 2018-11-02
gh-pages now builds successfully

Jakobo:
This addresses one half of #183. We needed to both (1) be able to generate the gh-pages branch again and (2) be able to map eyeglass.rocks to the new `A` record from github. When attempting a rebuild with node v8 and v10, the following problems were discovered:

* `gulp-gh-pages` uses an outdated version of `gift`
* `eyeglass-dev-site-builder` needs updated `eyeglass` & `node-sass`

This was fixed by changing the dependency on `gulp-gh-pages` to the community module `gulp-gh-pages-will`. To fix the eyeglass dependency, the values in `package-lock.json` were manually increased to a version of `eyeglass` and `node-sass` that do not run into an Abort Trap on OSX.


stefanpenner:
Awesome, thank you!

Issue #198, opened 2019-02-07
Extract sass type definitions to @types/node-sass.


chriseppstein:
The type definition for node-sass required changes. We need to combine the types in SassImplementation with the main node-sass types and to reconcile this against the type definitions generated for dart-sass.

chriseppstein:
Note: `SassError` currently extends `Error`, it needn't define the `message` property again.

PR #202, opened 2019-02-11
[WIP] Node sass types

chriseppstein:
Addresses https://github.com/sass-eyeglass/eyeglass/issues/198


packages/eyeglass/src/modules/EyeglassModules.ts
```diff
@@ -150,15 +150,15 @@ export default class EyeglassModules {
 
     // otherwise, return the module reference
     return mod!;
```
stefanpenner: is the trailing `!` still required?

PR #203, opened 2019-02-12
Fixes #200, using \x00 as cacheKey separator

### LINKED ISSUE #200: Cache key delimiters
There's some potential concern about cacheKey delimiters here that may be an easy fix.
https://github.com/sass-eyeglass/eyeglass/pull/197#discussion_r254786071

Jakobo:
This addresses #200, using @stefanpenner's suggestion of `\x00` as a cache key separator. This preempts any issues with `:` in a key name. I made the change and reran the tests to ensure we didn't break anything as part of this PR.


chriseppstein:
@Jakobo Thanks for the patch!

PR #204, opened 2019-02-15
Update typings for node-sass, broccoli-eyeglass, ember-cli-eyeglass, switch to yarn, update linter infrastructure.

chriseppstein:
This PR addresses the typings for node-sass which are still local, but [have been submitted to the DefinitelyTyped repo](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33017) and also properly converts broccoli-eyeglass to Typescript. Also addressed in this PR is an issue with the eyeglass typings for the project's main export (which is a class that is also a function that returns something other than a new instance when invoked without `new`).


chriseppstein:
Note: The broccoli-eyeglass and ember-cli-eyeglass builds on master are currently passing because broccoli isn't using the version of eyeglass that came from the monorepo 😭 This PR makes broccoli-and ember-cli builds use the monorepo versions.

chriseppstein:
This PR switches to yarn to install npm modules and drops a dependency on eyeglass-dev-site-builder (for now) which dropped a transitive dependency on a really old version of node-sass. I've also stopped running our coverage reports in travis. Thanks to these changes our travis build now runs in 5 min instead of 30+ minutes.

PR #219, opened 2019-02-28
Don't use package.json files to locate a module from the directory structure.

chriseppstein:
This is more efficient because it avoids a lot of filesystem access.

This also works around an issue in ember-cli where package.json files
seem to go missing during the build.


stefanpenner:
Nice!

PR #220, opened 2019-03-07
Package.json: restore file dependencies to link dependencies

stefanpenner:
restore `file` dependencies to `link`, as `file` in yarn only copies assets rather than symlink which makes me constantly troll myself in development.

@chriseppstein I do understand `link` is non-standard, and maybe it's not worth changing back. Would love to hear your thoughts.


rwjblue:
Link is definitely valid for yarn, it’s just non-standard for npm. AFAICT, this repo is using yarn for all it’s local development so I would think link is totally fine there. 

Am I missing something?

chriseppstein:
I like the semantics of `link:` over `file:` and I'm happy to merge it in, but this change breaks the build and I'm not sure what about my ember config is wrong that's causing this to fail.

rwjblue:
The error (during `ember test`) is:

> Error while loading rule '@typescript-eslint/restrict-plus-operands': You have used a rule which requires parserServices to be generated. You must therefore provide a value for the "parserOptions.project" property for @typescript-eslint/parser.

Most _likely_ need to follow what it says [here](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin#usage):

> If you want to use rules which require type information, you will need to specify a path to your tsconfig.json file in the "project" property of "parserOptions".
> 
> ```json
> {
>   "parser": "@typescript-eslint/parser",
>   "parserOptions": {
>     "project": "./tsconfig.json"
>   },
>   "plugins": ["@typescript-eslint"],
>   "rules": {
>     "@typescript-eslint/restrict-plus-operands": "error"
>   }
> }
> ```

chriseppstein:
@rwjblue I tried adding parser options before I commented. it didn't fix the error. 🤷‍♂️

PR #223, opened 2019-03-10
Perf instrumentation

chriseppstein:
Instrument eyeglass and broccoli-eyeglass for performance metrics gathering with heimdalljs.

This introduces metrics collections for the following:

#### Always enabled
* `eyeglass:broccoli:build:compileTree:<directory>` - Each broccoli tree that gets compiled by broccoli-eyeglass.
  * Additional Metrics collected per tree:
    * `cacheMissCount` - how many sass files had to be recompiled
    * `cacheHitCount`- how many sass files were cached
    * `numSassFiles` - the number of sass files.
    * `nodeSassTime` - The time node-sass reports spending on compilation (includes waiting in the libuv worker queue).
    * `importCount` - The number of `@import` statements encountered.
    * `uniqueImportCount` - The number of `@import` statements that couldn't be deduplicated.
  * eyeglass:broccoli:build:invalidation - the cost of invalidating cache results based on what's changed since the last build.
* eyeglass:instantiation - The cost of instantiation and module creation.
  * `eyeglass:modules` - eyeglass modules creation during instantiation
    * `eyeglass:modules:discovery` - discovery of eyeglass modules within npm modules.
    * `eyeglass:modules:resolution` - resolving transitive eyeglass modules into a single version.

#### Conditionally enabled

When the environment variable `EYEGLASS_PERF_DEBUGGING` is set to a non-empty string additional perf metrics are collected. These incur performance overhead and can bloat the
heimdall data returned.

Heimdall nodes are added for:
* eyeglass:import:module - the cost of module import (most imports are these)
  * eyeglass:import:assets - the cost of asset importing (delegated from the module importer)
    * eyeglass:import:fs - the cost of fs importing (delegated from the asset importer)

A heimdall monitor named `sassFns` is enabled causing stats to be collected against the broccoli-eyeglass tree node:
* The key is the sass function name that was invoked and the value is an object with a `count` and a `time` similar to how the heimdall `fs` monitor works. these times are collected with nanosecond precision. This only monitors sass functions implemented in js.

### DIFF
--- a/packages/eyeglass/src/Eyeglass.ts
+++ b/packages/eyeglass/src/Eyeglass.ts
@@ -14,6 +14,7 @@ import { SassFunction } from "node-sass";
 import { SassImplementation, helpers as sassHelpers } from "./util/SassImplementation";
 import { AsyncImporter } from "node-sass";
 import { UnsafeDict } from "./util/typescriptUtils";
+import heimdall = require("heimdalljs");
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const pkg: PackageJson = require("../package.json");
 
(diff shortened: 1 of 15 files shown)


packages/eyeglass/src/Eyeglass.ts
```diff
@@ -26,39 +27,46 @@ export default class Eyeglass implements IEyeglass {
...
-      for (let assetSource of this.options.eyeglass.assets.sources) {
-        this.assets.addSource(assetSource.directory, assetSource);
+      // auto-add asset paths specified via options
+      if (this.options.eyeglass.assets.sources) {
+        for (let assetSource of this.options.eyeglass.assets.sources) {
+          this.assets.addSource(assetSource.directory, assetSource);
+        }
       }
+    } catch(e) {
+      // typescript needs this catch & throw to convince it that the instance properties are initialized.
```
stefanpenner: typescript doesn't nicely support `try/finally` without `catch` ?

PR #226, opened 2019-03-21
Optimize eyeglass asset installation.

chriseppstein:
Eyeglass assets are installed for each invocation of `asset-url()` which can add performance overhead at scale especially when the same asset is installed many times.
    
`ember-cli-eyeglass` now intercepts asset installation and records which assets are installed to which locations. It does this by writing all those assets into a single tree for the entire app and all addons and engines after they are built, which it only needs to do once per distinct url. To accomplish this, I wrote a new broccoli-plugin in [this commit](https://github.com/linkedin/eyeglass/commit/5a048a9225bbcbbe86e051c6de5d63dbf318dfa3) which makes in possible to symlink arbitrary files into an output tree.
    
This PR has the benefit of also ensuring that asset urls can be resolved to any path instead of forcing those urls to resolve to a location relative to the current tree, which fixes a long-standing annoyance that forced assets to be duplicated in engines (note: this patch doesn't change the asset installation locations -- a custom resolver is required to accomplish this for now).

This patch was made more tricky because of my bone-headed insistence of having a generic broccoli sass plugin as the base class for the broccoli eyeglass plugin. The pure sass plugin is in charge of the caching strategy so it uses events to delegate knowledge of how additional outputs are restored from cache to the external code that is invoking this plugin and doing weird things with assets outside the output tree.

A few other things to note:

* ember-cli-eyeglass addons (and the broccoli eyeglass compiler plugin) have a slightly different annotation now (it will show the parent app as well as the addon/engine name). If there's code that's relying on the annotation names, it may break or cause reporting inconsistencies.
* When persistent caching is enabled, it now automatically sets up any addons or engines to use a different persistent cache key instead of requiring each one to set their cache key in the configuration.
* The format of the output cache data has changed... as such the cache key for that data has a version which has been incremented to forcibly invalidate all the old caches in the wrong format.
* The old handling for non-css assets in ember apps has been removed because this new approach supersedes it.

TODO: I need to add some additional test cases.


stefanpenner:
@chriseppstein nice, I plan to take a thorough look this evening/weekend.

packages/broccoli-eyeglass/src/index.ts
```diff
@@ -175,6 +151,26 @@ class EyeglassCompiler extends BroccoliSassCompiler {
     if (this.configureEyeglass) {
       this.configureEyeglass(eyeglass, options.eyeglass.engines.sass, details);
     }
+
+    // set up asset dependency tracking
+    eyeglass.assets.resolver((filepath, fullUri, realResolve, cb)  => {
+      this.events.emit("dependency", filepath).then(() => {
+        realResolve(filepath, fullUri, cb);
+      }, cb);
```
stefanpenner: if `realResolve` throws the `cb` errback wont be notified, is that intentional?

I would expect:

```js
then(() => realResolve(filePath, fullUri, cb)).catch(cb);
```


An idea for the future, improving the eyeglass apis to natively support promises could result in:

```js
eyeglass.assets.resolver(async (filepath, fullUri) => {
  await this.events.emit('dependency');
  return realResolve(filepath, fullUri);
});

File: packages/broccoli-eyeglass/src/index.ts [hunk 2]
```diff
@@ -175,6 +151,26 @@ class EyeglassCompiler extends BroccoliSassCompiler {
...
+    });
+
+    eyeglass.assets.installer((file, uri, realInstall, cb) => {
+      realInstall(file, uri, (error: unknown, destFile?: string) => {
+        if (error) {
+          cb(error, file);
+        } else {
+          this.events.emit("additional-output", destFile, uri, file).then(() => {
+            cb(null, file);
+          }, cb);
```
stefanpenner: ^

packages/broccoli-eyeglass/src/types/sync-disk-cache.d.ts
```diff
@@ -1,16 +1,32 @@
...
+  key: undefined;
+  value: undefined;
+}
+
+interface SyncDiskCacheOptions {
+  location?: string;
+  compression?: 'deflate' | 'deflateRaw' | 'gzip';
+}
+
 declare class SyncDiskCache {
```
stefanpenner: left and issue, to simply port sync-disk-cache to typescript so that it brings it's own typings with: https://github.com/stefanpenner/sync-disk-cache/issues/17 (if you have free time, I would welcome the PR)

packages/ember-cli-eyeglass/src/broccoli-ln-s.ts
```diff
@@ -0,0 +1,82 @@
...
+
+type BroccoliSymbolicLinkerOptions =
+  Pick<BroccoliPlugin.BroccoliPluginOptions, "annotation" | "persistentOutput">;
+
+/**
+ * Creates symlinks to the source files specified.
+ * 
+ *  BroccoliSymbolicLinker
+ */
+export class BroccoliSymbolicLinker extends BroccoliPlugin {
```
stefanpenner: this plugin likely needs to be tested, https://github.com/broccolijs/broccoli-test-helper can be quite helpful for that.

packages/ember-cli-eyeglass/src/broccoli-ln-s.ts
```diff
@@ -0,0 +1,82 @@
...
+  /**
+   * Returns the number of symlinks that will be created.
+   */
+  numberOfFiles(): number {
+    return Object.keys(this.files).length;
+  }
+  /**
+   * Create the symlinks. Directories to them will be created as necessary.
+   */
+  build(): void {
```
stefanpenner: As implemented, I believe this will for each rebuild rebuild it's output entirely. This may result downstream plugins into having to do extra work, as they assume their input has changed.

One potentially simple solution, is to skip the work if there are no changes in `this.files` since last time.

packages/ember-cli-eyeglass/src/broccoli-ln-s.ts
```diff
@@ -0,0 +1,82 @@
...
+   * invoked.
+   *
+   * @param src The file that should be symlinked into the tree.
+   * @param dest the relative path from the tree's root to the location of the
+   *   symlink. the filename does not have to be the same.
+   * @returns the absolute path to the location where the symlink will be created.
+   */
+  // eslint-disable-next-line @typescript-eslint/camelcase
+  ln_s(src: string, dest: string): string {
+    this.files[dest] = src;
```
stefanpenner: do we care if the `ln_s` is invoke twice with the same `dest` but differing `src` ? Is the rule simply last write wins?

packages/ember-cli-eyeglass/src/index.ts
```diff
@@ -77,10 +83,72 @@ function localEyeglassAddons(addon): Array<{path: string}> {
...
+  isApp: boolean;
+  app: any; // is this safe to cache across builds?
+}
+interface EyeglassAppInfo {
+  assets: BroccoliSymbolicLinker;
+  sessionCache: Map<string, string | number>;
+}
+
 const EMBER_CLI_EYEGLASS = {
   name: 'ember-cli-eyeglass',
```
stefanpenner: convention is to do `require('path/to/own/package.json').name`

packages/ember-cli-eyeglass/src/index.ts
```diff
@@ -77,10 +83,72 @@ function localEyeglassAddons(addon): Array<{path: string}> {
...
+    let parentPath = this.parent.root;
+    if (isApp) {
+      apps.push(app);
+      // we create the symlinker in persistent mode because there's not a good
+      // way yet to recreate the symlinks when sass files are cached. I would
+      // worry about it more but it seems like the dist directory is cumulative
+      // across builds anyway.
+      EYEGLASS_INFO_PER_APP.set(app, {
+        sessionCache: new Map(),
+        assets: new BroccoliSymbolicLinker({}, {annotation: app.name, persistentOutput: true})
```
stefanpenner: Making `persistentOutput` a parameter of `BroccoliSymbolicLinker` then requires `BroccoliSymbolicLinker`'s `build` method conditionally handling it's own cleanup. Currently it does not, and likely results in links never being removed, only added.

packages/ember-cli-eyeglass/src/index.ts
```diff
@@ -77,10 +83,72 @@ function localEyeglassAddons(addon): Array<{path: string}> {
...
   postBuild(result) {
     Eyeglass.resetGlobalCaches();
+    for (let app of apps) {
+      let appInfo = EYEGLASS_INFO_PER_APP.get(app);
+      if (appInfo) {
+        appInfo.assets.reset();
+        appInfo.sessionCache.clear();
+      } else {
+        // eslint-disable-next-line no-console
+        console.warn("eyeglass app info is missing during postBuild.");
```
stefanpenner: Is this actionable for the user? If so, we should likely use `this.ui` which should be an instance of https://github.com/ember-cli/console-ui and facilitates a way for unified output.

If not, then we should likely not output it by default, rather then debug logging.

packages/ember-cli-eyeglass/src/index.ts
```diff
@@ -77,10 +83,72 @@ function localEyeglassAddons(addon): Array<{path: string}> {
...
+      }
+    }
+  },
+  postprocessTree(type, tree) {
```
stefanpenner: @rwjblue i suspect this needs to implement cacheKeyForTree, can you confirm?

packages/ember-cli-eyeglass/src/broccoli-ln-s.ts
stefanpenner: Further reading suggests yes, there is likely some missing book-keeping here.

packages/broccoli-eyeglass/src/broccoli_sass_compiler.ts
```diff
@@ -644,23 +696,38 @@ export default class BroccoliSassCompiler extends BroccoliPlugin {
...
+        "firing 'cached-asset' for each asset url for %s: %s",
+        details.sassFilename,
+        allUrls.join(", ")
+      );
+    }
+    for (let url of allUrls) {
+      let sourceFile = urls[url];
+      eventPromises.push(this.events.emit("cached-asset", sourceFile, url))
+    }
+    return RSVP.all(eventPromises);
```
stefanpenner: do we want to fast fail, if any of the promises reject? Or should we wait until al the work is done, before proceeding?

stefanpenner:
@chriseppstein I think the overall approach here is very interesting. Good way to have a single "install" step.

packages/broccoli-eyeglass/src/types/sync-disk-cache.d.ts
chriseppstein: you don't have to port to typescript to provide typings. you can just publish this type definition with your project and keep it up to date when the public API changes.

packages/ember-cli-eyeglass/src/broccoli-ln-s.ts
chriseppstein: The tree is reset in postBuild causing `this.files` to become empty.

packages/ember-cli-eyeglass/src/broccoli-ln-s.ts
chriseppstein: It would be a breaking change to eyeglass to do anything other than maintain the behavior of having the last write win.

packages/ember-cli-eyeglass/src/index.ts
chriseppstein: Yes, you're correct. To remove those files requires tracking if every url in every addon that added it was removed. If the source changes in a way that would cause the image to no longer be installed, then the reference to that image is also removed. I know that means there's an image still floating around in the output directory until the next restart but it's a ton of work to fix a problem that basically isn't an issue in practical usage because there's not stray urls floating around to the output location in the authored source code.

packages/ember-cli-eyeglass/src/index.ts
chriseppstein: This doesn't happen anymore. I'll remove it. But I'm glad to learn about `this.ui`, there's probably other places in my code that should be using it.

packages/broccoli-eyeglass/src/broccoli_sass_compiler.ts
chriseppstein: Both of those behaviors are fine in this situation.

packages/broccoli-eyeglass/src/types/sync-disk-cache.d.ts
stefanpenner: Yup you are correct, but I have been finding the TS port sorta cathartic :P Good evening while watching TV exercise.  Plus I can write fancy ES2018 syntax, for old nodes.

packages/ember-cli-eyeglass/src/broccoli-ln-s.ts
stefanpenner: Yes, but combined with `persistentOutput: true`, the contents `this.outputPath`  may not realize deletions.

chriseppstein: You're correct. It won't delete. I decided to not write that code. As I explained in a different comment, the cost of doing that is a lot of complicated code and bookkeeping and the risk of leaving those files in place as a stale cache is very low in practice because there won't be any dangling references to the asset in the code we produce.

packages/ember-cli-eyeglass/src/index.ts
chriseppstein: From the docs for `cacheKeyForTree`:

```
This method is invoked prior to calling treeFor with the same tree name.

You should override this method if you implement custom treeFor or treeFor* methods, which cause addons to opt-out of this caching.
```

Based on that, it seems like adding this method doesn't help for this addon.

packages/broccoli-eyeglass/src/index.ts
chriseppstein: The contract for resolve is to report all errors via the callback, so it shouldn't throw. Is this just out of extreme caution for code that doesn't comply with the contract?

PR #229, opened 2019-03-28
Smart js invalidation

chriseppstein:
We were using hash-for-dep to invalidate the cache if any javascript in an eyeglass module changed. But many eyeglass addons are hybrid packages that have a lot of javascript that isn't relevant to the sass files.

This patch switches to creating a list of dependencies of js files starting with the eyeglass exports file and only considers those files as relevant to the sass file cache. This approach is slower because it
parses the js files for import and require statements, but because it is a more narrow cache, that extra cost pays for itself.

This only affects eyeglass modules marked as inDevelopment which use an eyeglass exports file.


packages/broccoli-eyeglass/src/broccoli_sass_compiler.ts
```diff
@@ -627,26 +627,33 @@ export default class BroccoliSassCompiler extends BroccoliPlugin {
...
+    let stat = statSync(file);
+    let key;
+    if (!isRealPath && stat.isSymbolicLink) {
+      key = this.fileKey(realpathSync(file), true);
     } else {
-      let data = fs.readFileSync(absolutePath, "UTF8");
-      let hash = crypto.createHash("md5").update(data).digest("hex");
-      this.buildCache.set(cacheKey, hash);
-      return hash;
+      key = `${stat.mtimeMs}:${stat.size}:${stat.mode}`;
```
stefanpenner: I think using `mtimeMs` is a breaking change, as it was introduced in node 8.10 https://github.com/nodejs/node/commit/157fc28710178e328bd6d3a66c33e2649ba81155#diff-acabf706a8aa070a8796e3573f7e4678R556 but this and ember-cli-eyglass [still support node >= 6](https://github.com/linkedin/eyeglass/blob/b130d929d28a3ffcb5a688a855f2a5cab6393e64/packages/broccoli-eyeglass/package.json#L63).

I think it's good to use, and may justify the major bump. Especially as Node 6 is technically supposed to EOL shortly.

Other upsides of that would be that you can start using native `async/await` in your code.

So options as i see:

* keep support for node 6, and fallback to `mtime` if `mtimeMs` is missing
* drop support for node 6, and do a major bump. (node 6 is EOL shortly anyways)

packages/broccoli-eyeglass/src/broccoli_sass_compiler.ts
```diff
@@ -627,26 +627,33 @@ export default class BroccoliSassCompiler extends BroccoliPlugin {
...
+  }
+
+  /* compute a key for a file that will change if the file has changed. */
+  fileKey(file: string, isRealPath = false): string {
+    let cachedKeyKey = `fileKey(${file})`;
+    let cachedKey = this.buildCache.get(cachedKeyKey) as string;
+    if (cachedKey) { return cachedKey; }
+    let stat = statSync(file);
+    let key;
+    if (!isRealPath && stat.isSymbolicLink) {
```
stefanpenner: this conditional may not be correct, I believe`stat.isSymbolicLink`  is a method, not a boolean. Are you intending to check for the existence of the the method, of if the stat itself is a link?

stefanpenner: Also, I believe `statSync` [by default follows symlinks](http://man7.org/linux/man-pages/man2/fstat.2.html), given that when would the result of `statSync(xyz).isSymbolicLink() === true`?

packages/broccoli-eyeglass/src/broccoli_sass_compiler.ts
```diff
@@ -627,26 +627,33 @@ export default class BroccoliSassCompiler extends BroccoliPlugin {
...
-      return cachedHash as string;
+    return this.fileKey(absolutePath);
+  }
+
+  /* compute a key for a file that will change if the file has changed. */
+  fileKey(file: string, isRealPath = false): string {
+    let cachedKeyKey = `fileKey(${file})`;
+    let cachedKey = this.buildCache.get(cachedKeyKey) as string;
+    if (cachedKey) { return cachedKey; }
+    let stat = statSync(file);
```
stefanpenner: statSync can throw if the value has gone missing, or if it is a broken symlink. Should we handle this, or should these scenarios crash the build?

packages/broccoli-eyeglass/src/index.ts
```diff
@@ -241,6 +240,28 @@ class EyeglassCompiler extends BroccoliSassCompiler {
...
+    let jsPath = mod.mainPath;
+    if (!jsPath) {
+      return ""
+    }
+    let cacheKey = `hashForJs(${jsPath})`;
+    let cachedHash = this.buildCache.get(cacheKey) as string;
+    if (cachedHash) {
+      return cachedHash;
+    }
+    let files = dependencyTree.toList({filename: jsPath, directory: path.dirname(jsPath)});
```
stefanpenner: As the order of `toList` here will affect the `hash`, do you know it is stable and will not spuriously produce differing hashes. If so awesome, if not you may want to lexicographically sort before hashing.

packages/broccoli-eyeglass/src/broccoli_sass_compiler.ts
chriseppstein: @stefanpenner Thanks for pointing this out. I'll write a shim that uses it if available.

stefanpenner: @chriseppstein perfect. I suspect other parts of the broccoli/build ecosystem would be better served by also switching to `mtimeMs`. So thanks for surfacing this, I totally wasn't aware it was a feature.

Once node 8 is dead, I'll explore utilize it across the board.

packages/broccoli-eyeglass/src/broccoli_sass_compiler.ts
chriseppstein: That explains the infinite loop I got lol.

chriseppstein: Evidently never. I've removed the symbolic link handling. 🎉

packages/broccoli-eyeglass/src/broccoli_sass_compiler.ts
chriseppstein: I'll add some error handling for this case, we shouldn't crash.

packages/broccoli-eyeglass/src/index.ts
chriseppstein: Good question. According to the documentation, this is a topological sort of the dependency order based on how the files are discovered in the source files. My understanding is that it is stable.

PR #248, opened 2020-05-28
Simplify URI.system()

mikrostew:
This simplifies what URI.system() does to optimize its performance, since it is called many times (on the order of 800,000 times in our codebase).

This is what is used to do:
* create a URI object
* run the input uri through a regex to split off the base path, query string, and hash string
* convert the path to use the system path separator
* normalize the path
* convert the path to use the standard "/' separator
* convert the path to use the system path separator

Because this function is only called with file and directory paths, and contains some duplication of effort, it can be reduced to doing this:
* convert the input uri to use the system path separator (`path.sep`)
* normalize the path

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,5 +4,6 @@ node_js:
 - '10'
 - '12'
 install: yarn
-allow_failures:
-- node_js: '8'
+jobs:
+  allow_failures:
+  - node_js: '8'
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,5 @@ node_js:
 - '10'
 - '12'
 install: yarn
+allow_failures:
+- node_js: '8'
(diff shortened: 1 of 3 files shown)


mikrostew:
Hmm, looks like [this failed on Node 12](https://travis-ci.org/github/linkedin/eyeglass/builds/692352217) in `ember-cli-eyeglass` due to a build error with babel
```
lerna ERR! yarn run test stderr:
Build Error (broccoli-persistent-filter:Babel > [Babel: ember-asset-loader]) in ember-asset-loader/test-support/loaded-asset-state.js
No "exports" main resolved in /home/travis/build/linkedin/eyeglass/node_modules/@babel/helper-compilation-targets/package.json
```

chriseppstein:
Cherry-picked to current release as a81e356ebf6883fe874dbeae358e4ca161c7df46.

chriseppstein:
Thank you for the patches!

Released as dist-tag `v2`:
 - broccoli-eyeglass@5.5.1
 - ember-cli-eyeglass@6.5.1
 - eyeglass@2.5.1

Released as dist-tag `current`:
 - broccoli-eyeglass@6.0.1
 - ember-cli-eyeglass@7.0.1
 - eyeglass@3.0.1

PR #265, opened 2021-01-28
Should execute after ember-css-modules

timlindvall:
ember-css-modules should run its build step before ember-cli-eyeglass. This allows for CSS Modules intermediate builds to work nicely with Eyeglass. (See https://github.com/salsify/ember-css-modules/blob/master/docs/PREPROCESSORS.md)


chriseppstein:
Thanks for the fix!

PR #279, opened 2021-09-16
feat: Add config for conditional Embroider builds.

timlindvall:
BREAKING BUGFIX

Requires apps to declare if they are using Embroider to build
via config option in `ember-cli-build.js`.

## More Details

Apps may opt to build using Embroider conditionally. This means they'll have `@embroider/core` included in their dependencies but may _not_ be building using Embroider. This can cause extraneous files to end up in the output tree, which can cascade into failures in downstream addons (such as `ember-cli-terser`).

This adds a config that apps **must** use to declare if they're using Embroider conditionally. This solely controls whether `ember-cli-eyeglass` builds using "Embroider mode" or not. (The previous code of detecting what dependencies the app has added is removed.)

## How to Reproduce the Bug

1. In `eyeglass-test-app`, add `ember-cli-terser` and `@embroider/core` to the build. Do _not_ make any changes to `ember-cli-build.js`.
2. Add a Glimmer component to the addon. Be sure to use at least one decorator, such as `@tracked` or `@action`.
3. Run `ember build --prod`.
4. `ember-cli-terser` will fail the build when it trips across the uncompiled decorator in the code.

If you run with `BROCCOLI_DEBUG="*"`, you'll note a bunch of files in the `output` dir for `ember-cli-eyeglass` that shouldn't be there. These eventually end up in the `input` to `ember-cli-terser`.

## How to Verify Fix

1. In `eyeglass-test-app`, after applying the above steps, add `embroiderEnabled: false` to the eyeglass config in `ember-cli-build.js`.
2. Run `ember build --prod`.
3. It should succeed! 🎉 

When running with Broccoil Debug, the `output` dir should only include compiled CSS files and other CSS files that didn't need to be compiled, nested in an `assets` dir.

I also verified this fix by running `yarn test` on `ember-cli-eyeglass`.


packages/ember-cli-eyeglass/README.md
```diff
@@ -154,6 +154,35 @@ The contents of `my-engine/addon/styles/my-addon.scss` will be added to `dist/en
 The contents of `my-engine/addon/styles/secondary.scss` will be added to `dist/engine-dist/my-engine/assets/engine.css`
 The contents of `my-engine/addon/styles/_shared.scss` will only be included if `my-addon.scss` or `secondary.scss` explicitly import them.
 
+### Embroider Support
+
+This addon works with Embroider builds as well as classic Ember CLI builds. By default, the addon will determine if your app is using Embroider by checking for the presence of `@embroider/core` in the `dependencies` or `devDependencies` listed in `package.json`.
+
+If you have added code to `ember-cli-build.js` that _conditionally_ uses Embroider to build your application, you can provide the `embroiderEnabled` configuration option to declare to `ember-cli-eyeglass` whether you are using Embroider for this build.
```
thousand, line 161: Would addons needing to support Embroider have to do anything for this?

packages/ember-cli-eyeglass/src/index.ts
```diff
@@ -77,19 +77,20 @@ function isUsingEmbroider(app: any): boolean {
   return !!(dependencies["@embroider/core"] || devDependencies["@embroider/core"]);
 }
 
-function embroiderEnabled(): boolean {
-  return g.EYEGLASS.projectInfo.usingEmbroider;
+//eslint-disable-next-line @typescript-eslint/no-explicit-any
+function embroiderEnabled(config: any): boolean {
+  return typeof config.embroiderEnabled !== 'undefined' ? config.embroiderEnabled : g.EYEGLASS.projectInfo.usingEmbroider;
```
thousand, line 82: seems like `??` would be better than the ternary here.

packages/ember-cli-eyeglass/README.md
timlindvall: The app controls the build, so I don't think addons that are depended on by a conditional Embroider app would need to do anything.

thousand (approved):
LGTM! @rwjblue or @stefanpenner might want to take a look too?

```
              |    |    |
             )_)  )_)  )_)
            )___))___))___)\
           )____)____)_____)\\
         _____|____|____|____\\\__
---------\                   /---------
  ^^^^^ ^^^^^^^^^^^^^^^^^^^^^
    ^^^^      ^^^^     ^^^    ^^
         ^^^^      ^^^
```

packages/ember-cli-eyeglass/src/index.ts
```diff
@@ -77,19 +77,20 @@ function isUsingEmbroider(app: any): boolean {
   return !!(dependencies["@embroider/core"] || devDependencies["@embroider/core"]);
 }
 
-function embroiderEnabled(): boolean {
-  return g.EYEGLASS.projectInfo.usingEmbroider;
+//eslint-disable-next-line @typescript-eslint/no-explicit-any
+function embroiderEnabled(config: any): boolean {
+  return config.embroiderEnabled ?? g.EYEGLASS.projectInfo.usingEmbroider;
```
stefanpenner, line 82: If we go down the route this PR recommends, should we have both: 'isUsingEmbroider' dependency check, and this flag?

stefanpenner: The old flag being renamed to `hasEmbroiderDependencies` could help with the above mentioned ambiguity.

packages/ember-cli-eyeglass/README.md
```diff
@@ -154,6 +154,35 @@ The contents of `my-engine/addon/styles/my-addon.scss` will be added to `dist/en
...
+const USE_EMBROIDER = false;
+
+module.exports = function(defaults) {
+  let app = new EmberApp(defaults, {
+    eyeglass: {
+      discover: true,
+      /**
+       * Should be a boolean, true if using Embroider, false otherwise.
+       */
+      embroiderEnabled: USE_EMBROIDER,
```
stefanpenner, line 173:
```

**Comment by stefanpenner** [comment]
I left some notes, I think in general this is the right path forward in the short term.

@rwjblue should we unify the "conditional embroider" usage with https://github.com/embroider-build/embroider/tree/master/packages/test-setup ?

File: packages/ember-cli-eyeglass/src/index.ts (hunk 3 above)
Comment (timlindvall) [reply packages/ember-cli-eyeglass/src/index.ts:82]: I'd be worried about dropping the current behavior since that would be backwards-incompatible. If there's a way we can do this detection in a more reliable way, I'd prefer that over asking the app to declare it manually, since then apps wouldn't need to make any changes.

I added some info on this behavior in the readme, let me know if the description is sufficient. 👍

File: packages/ember-cli-eyeglass/README.md (hunk 4 above)
Comment (timlindvall) [reply packages/ember-cli-eyeglass/README.md:176]: Fixed!

File: packages/ember-cli-eyeglass/src/index.ts (hunk 3 above)
Comment (timlindvall) [reply packages/ember-cli-eyeglass/src/index.ts:82]: Renamed `isUsingEmbroider` to `hasEmbroiderDependencies`. (Sorry, didn't show me your second comment until I did a hard refresh. 🤦 )

**Review by stefanpenner — CHANGES_REQUESTED** [review]

File: packages/ember-cli-eyeglass/src/index.ts [hunk 5]
```diff
@@ -70,26 +70,27 @@ function isLazyEngine(addon: any): boolean {
 }
 
 //eslint-disable-next-line @typescript-eslint/no-explicit-any
-function isUsingEmbroider(app: any): boolean {
+function hasEmbroiderDependency(app: any): boolean {
```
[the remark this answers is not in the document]
timlindvall: Sounds good. Think it's worth keeping the dependency checks and adding a warning if the addon detects a de-sync? (e.g.: The app has a dep on `embroider/core` but hasn't set the flag)

I'm considering this because the downstream build errors that you could get from this being misconfigured can be cryptic and difficult to trace back to `ember-cli-eyeglass`.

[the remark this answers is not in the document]
timlindvall: nvm, I realized that this check wouldn't actually catch the situation I'm solving with this PR.

Alright, updated diff to remove the auto-detect code.

stefanpenner (approved):

stefanpenner:
@rwjblue WDYT?

* do we need a major, or is the embroider support change ok as a minor
* are you ok with the flag change?

rwjblue (approved):
IMHO, this is at worst a "breaking bugfix" (since the previous Embroider detection system introduced bugs for folks that had the packages but weren't using Embroider) so this is fine without a major.

PR #283, opened 2021-10-20
fix: Use prev. asset bundling code if Classic.

timlindvall:
7.0.3 introduced changes to the asset bundling code to support Embroider
builds. At the time bundling the assets was conditional based on whether
Classic or Embroider was being used. This changed in 7.0.6, since in
Embroider there was a timing issue with assets sometimes not showing up
in the final artifact.

However, this appears to have caused another issue where extra assets
that shouldn't be present are showing up in the output tree. To fix
this, this commit re-introduces the asset code that was removed in 7.0.3
and conditionally picks what asset build pipeline to use based on
whether the app is being built using Classic or Embroider.

## Testing Done

- Tested an addon (using `ember serve` classic build) that depends on multiple packages with Eyeglass assets. Used `broccoli-debug` to manually check the output of the `ember-cli-eyeglass` step, as well as the contents of the `dist` folder, when:
    - Building using 7.0.4.
    - Building using 7.1.0 (without this fix). (Has extra assets in the output dir.)
    - Building using a local copy of `ember-cli-eyeglass` with this fix. (Doesn't have extra assets in the output dir.)
- Ran `yarn test` locally on `ember-cli-eyeglass` and verified there were no failures reported.

### DIFF
--- a/packages/ember-cli-eyeglass/src/index.ts
+++ b/packages/ember-cli-eyeglass/src/index.ts
@@ -26,10 +26,11 @@ interface EyeglassAddonInfo {
   parentPath: string;
   isApp: boolean;
   app: any;
-  assets: BroccoliSymbolicLinker;
+  assets?: BroccoliSymbolicLinker;
 }
 
 interface EyeglassAppInfo {
+  assets?: BroccoliSymbolicLinker;
   sessionCache: Map<string, string | number>;
 }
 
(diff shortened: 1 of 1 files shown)


thousand (approved):
I think this should probably include new or updated tests to prevent future churn/regression on this. But in the interest of unblocking other work, i think it'd be fine to add that in a future rev.

oa495:
Was this published in a version?

